PR #7, opened 2018-12-05
Default to Jaeger username 'auth'

keitwb:


dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/TracerInstaller.java
```diff
@@ -14,6 +14,10 @@ public static synchronized void installGlobalTracer() {
     if (!io.opentracing.util.GlobalTracer.isRegistered()) {
       Configuration conf = Configuration.fromEnv();
 
+      if (Strings.isNullOrEmpty(conf.getReporter().getSenderConfiguration().getAuthUsername())) {
+        conf.getReporter().getSenderConfiguration().withAuthUsername("auth");
```
mpetazzoni: Shouldn't this only happen if a token/password is also set? Or is the sender correctly ignoring authentication when no token/password is configured?

keitwb: It does but I will conditionalize this to make it cleaner.

PR #9, opened 2018-12-21
Revert split between tooling and instrumentation

keitwb:
This brings us back to closer alignment with DD's agent.

I also moved the instrumentation utils helper to the dd-trace-ot package.


rmfitzpatrick:
What is the intention of this?  iirc we wanted to have the url util in instrumentation to not force the dependency and because it's only used instrumentations only.

PR #13, opened 2019-01-23
Add server span kind to jax-rs-controller spans

rmfitzpatrick:
For uninstrumented servers, having the first traced operation at the controller level use this span kind will be informative.


mpetazzoni:
As is I'm a -1 on this until we understand the impact more. It would lead to us considering both the `jax-rs` and the `jax-rs-controller` spans as "endpoints" of the service, which will be misleading.

We should also test this thoroughly to make sure it doesn't lead to weird interactions with how we determine services and interactions between services, etc.

PR #14, opened 2019-01-23
Create root span in jax-rs-annotation if none exists

rmfitzpatrick:
For cases where jax-rs utilizing frameworks aren't instrumented and no active span has been created for the traced request handler, here we create one to parent the jax-rs-controller spans so they aren't used as a root span.

Instrumentation tests have been updated to not rely on runUnderTrace's active span.


dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
```diff
@@ -105,9 +107,20 @@ public static Scope nameSpan(@Advice.Origin final Method method) {
...
 
-      final Scope scope = GlobalTracer.get().scopeManager().active();
-      if (scope != null && !resourceName.isEmpty()) {
-        Span span = scope.span();
+      final List<Scope> scopeList = new ArrayList<>();
+      Scope parentScope = GlobalTracer.get().scopeManager().active();
+      if (parentScope == null) {
+        parentScope =
+            GlobalTracer.get()
+                .buildSpan("JAX-RS")
```
mpetazzoni: You should default to naming this with the `operationName` to make sure that, even if `resourceName` is empty, we don't end up with spans named the same for all endpoints. Remember this impacts metrics baselines.

dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
```diff
@@ -63,7 +65,7 @@ public JaxRsAnnotationsInstrumentation() {
   public static class JaxRsAnnotationsAdvice {
 
     @Advice.OnMethodEnter(suppress = Throwable.class)
-    public static Scope nameSpan(@Advice.Origin final Method method) {
+    public static List<Scope> nameSpan(@Advice.Origin final Method method) {
```
mpetazzoni: You should return a `Stack<Scope>` instead to clearly indicate that this is a stack that needs to be popped in order to close the scopes.

dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
```diff
@@ -105,9 +107,20 @@ public static Scope nameSpan(@Advice.Origin final Method method) {
...
+      final List<Scope> scopeList = new ArrayList<>();
+      Scope parentScope = GlobalTracer.get().scopeManager().active();
+      if (parentScope == null) {
+        parentScope =
+            GlobalTracer.get()
+                .buildSpan("JAX-RS")
+                .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
+                .withTag(Tags.COMPONENT.getKey(), "jax-rs")
+                .startActive(true);
+        scopeList.add(parentScope);
```
mpetazzoni: With a stack: `scopes.push(parentScope)`

dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
```diff
@@ -135,21 +148,27 @@ public static Scope nameSpan(@Advice.Origin final Method method) {
...
-          .buildSpan(operationName)
-          .withTag(Tags.COMPONENT.getKey(), "jax-rs-controller")
-          .startActive(true);
+      final Scope controllerScope =
+          GlobalTracer.get()
+              .buildSpan(operationName)
+              .withTag(Tags.COMPONENT.getKey(), "jax-rs-controller")
+              .startActive(true);
+
+      scopeList.add(0, controllerScope);
```
mpetazzoni: Again, with a stack: `scopes.push(controllerScope)`

dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
```diff
@@ -135,21 +148,27 @@ public static Scope nameSpan(@Advice.Origin final Method method) {
...
+      return scopeList;
     }
 
     @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
     public static void stopSpan(
-        @Advice.Enter final Scope scope, @Advice.Thrown final Throwable throwable) {
+        @Advice.Enter final List<Scope> scopes, @Advice.Thrown final Throwable throwable) {
       if (throwable != null) {
-        final Span span = scope.span();
+        final Span span = scopes.get(0).span();
```
mpetazzoni: `scopes.peek().span()`

dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
rmfitzpatrick: Great idea.  Updated.

dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
```diff
@@ -135,21 +122,54 @@ public static Scope nameSpan(@Advice.Origin final Method method) {
...
+                .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
+                .withTag(Tags.COMPONENT.getKey(), "jax-rs")
+                .startActive(true);
+        scopeStack.push(parentScope);
+      }
+
+      if (!resourceName.isEmpty()) {
+        Span span = parentScope.span();
+        span.setOperationName(resourceName);
+        Tags.SPAN_KIND.set(span, Tags.SPAN_KIND_SERVER);
```
mpetazzoni: I think we should always set those tags (kind, component, `http.url` and http.method`) even if resourceName is empty (and also avoid the duplication between here and line 132).

rmfitzpatrick: I went back and forth on this and thought the all-or-nothing was cleaner in some regards but think I agree overall.  Will update.

PR #17, opened 2019-03-06
Add SignalFx-specific API classes/interfaces

keitwb:


dd-trace-api/src/main/java/signalfx/trace/api/Trace.java
```diff
@@ -0,0 +1,20 @@
+package signalfx.trace.api;
```
rmfitzpatrick: Mind using `signalfx.tracing.api` for uniformity?

keitwb: Uniformity with what?  Do we have some other Java code out there with that package name?

rmfitzpatrick: consistency across instrumentation libraries (signalfx-tracing)

mpetazzoni: Any particular reason why this isn't under `com.signalfx.public.tracing.api` ?

README.md
```diff
@@ -13,16 +13,22 @@ as the OpenTracing `GlobalTracer` to enable additional custom instrumentation.
...
 $ java -javaagent:signalfx-tracing.jar <rest of command-line arguments...>
 ```
 
+This invocation assumes that you have a [locally deployed SignalFx Smart
+Agent](https://docs.signalfx.com/en/latest/apm/apm-deployment/smart-agent.html)
+accessable on `localhost`.  If you want to send traces directly to the Smart
+Gateway or if your Smart Agent is only accessible on a different host (e.g. in
+a containerized environment), set the trace host via the `SIGNALFX_AGENT_HOST`
+envvar to the approprate value (default is `localhost`).  You can override the
+entire trace target URL with the `SIGNALFX_AGENT_ENDPOINT` envvar (default:
```
achandras: Should this say `SIGNALFX_AGENT_ENDPOINT_URL`?

dd-trace-api/src/main/java/signalfx/trace/api/Trace.java
keitwb: We can do `com.signalfx.tracing.api` but the word `public` is not allowed in package names since it is a keyword in Java and each dot-separated component of a package name has to be a valid identifier: https://docs.oracle.com/javase/specs/jls/se8/html/jls-7.html#jls-7.4.1

README.md
keitwb: It should be SIGNALFX_ENDPOINT_URL per our discussions.

dd-trace-api/src/main/java/signalfx/trace/api/Trace.java
mpetazzoni: Ah yeah sorry the `public` is in our artifactId. Let's do `com.signalfx.tracing.api` then, if possible.

keitwb: ok, updated to that package name.

README.md
keitwb: fixed.

PR #21, opened 2019-04-01
Netty nonstandard status handling

rmfitzpatrick:
In cases where Netty users don't want automatic error tagging or trace collector error reporting for 5xx http statuses, these changes introduce the 
`signalfx.instrumentation.client.nonstandard.http.status.code.<code>=true`
and
`signalfx.instrumentation.server.nonstandard.http.status.code.<code>=true` properties to tag specified status codes for client and servers w/ `http.status.orig`.


dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
```diff
@@ -35,14 +37,14 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
         if (finishSpan) {
           Tags.ERROR.set(span, Boolean.TRUE);
           span.log(Collections.singletonMap(ERROR_OBJECT, throwable));
-          Tags.HTTP_STATUS.set(span, 500);
+          NettyUtils.setClientSpanHttpStatus(span, 500);
```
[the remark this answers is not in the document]
keitwb: And then the instrumentation simply puts whatever is the actual status code in the span tag?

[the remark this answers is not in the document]
rmfitzpatrick: Maybe it's not guaranteed the msg will have been able to determine a status if `fireChannelRead` throws and this is a sane default to alert?  Should I attempt that in a try block?

README.md
```diff
@@ -59,7 +59,7 @@ given in the table below.
 | _Java MDC_ | * | `-Dsignalfx.logs.injection=true` on Java invocation | Injects `signalfx.trace_id` and `signalfx.span_id` to MDC contexts |
 | **Mongo Client** | 3.1+ | `mongo` | |
 | **Mongo Async Client** | 3.3+ | `mongo` | |
-| **Netty Client and Server** | 4.0+ | `netty` | |
+| **Netty Client and Server** | 4.0+ | `netty` | Nonstandard HTTP status code tagging w/ ```-Dsignalfx.instrumentation.netty.<client or server>.nonstandard.http.status.<code>=true``` to circumvent Status5xxDecorator |
```
mpetazzoni: A generally accepted way to represent choice here is `-Dsignalfx.instrumentation.netty.{client,server}.nonstandard.http.status.<code>=true`

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/NettyUtils.java
```diff
@@ -0,0 +1,53 @@
...
+@Slf4j
+public class NettyUtils {
+
+  public static final String NETTY_REWRITTEN_SERVER_STATUS_PREFIX =
+      "instrumentation.netty.server.nonstandard.http.status.";
+  public static final String NETTY_REWRITTEN_CLIENT_STATUS_PREFIX =
+      "instrumentation.netty.client.nonstandard.http.status.";
+  public static final IntTag ORIG_HTTP_STATUS = new IntTag(Tags.HTTP_STATUS.getKey() + ".orig");
+
+  private static final Map<Integer, Boolean> serverStatusCache = new HashMap<>();
```
mpetazzoni: I'm not sure I understand the need for those caches? The `Config.getBooleanSettingFromEnvironment` will ultimately lead to a hash lookup as well (into the system properties map). Plus, since status is likely to be greater than 128, you'll be creating a new `Integer` instance for boxing every time you look up into those caches, so if this is meant for performance I'm not sure the benefit is clear vs pure simplicity.

rmfitzpatrick: I was operating under the assumption that lookup by integer would have been preferable to string concatenation, lookup, and casting to boolean.  Removing as it may be adding needless complexity.

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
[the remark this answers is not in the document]
rmfitzpatrick: @keitwb, this ok?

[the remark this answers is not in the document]
keitwb: I guess so, but I still don't understand why the instrumentation is tagging it as a 500 without it getting the status code directly from some response object that actually says it's a 500.  Seems like just setting the span to `error: true` is enough without trying to fake a status.

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
```diff
@@ -35,14 +37,22 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
         if (finishSpan) {
           Tags.ERROR.set(span, Boolean.TRUE);
           span.log(Collections.singletonMap(ERROR_OBJECT, throwable));
-          Tags.HTTP_STATUS.set(span, 500);
+          int status;
+          try {
```
mpetazzoni: Simpler to just to set the tag in the try:

```java
try {
  int status = ((HttpResponse) msg).getStatus().code();
  NettyUtils.setClientSpanHttpStatus(span, status);
} catch (final Throwable ex) {
  // Ignored.
}
```

Is the behavior change (compared to prior which always set 500) expected though?

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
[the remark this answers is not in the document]
rmfitzpatrick: I was referring to my updates, which no longer set the 500.

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
rmfitzpatrick: I wanted not to hide exceptions from the status setting utility, but agree it's simpler.  @keitwb and I agreed it's better not to fabricate a status code in the client instrumentation.

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/server/HttpServerResponseTracingHandler.java
```diff
@@ -28,12 +30,12 @@ public void write(final ChannelHandlerContext ctx, final Object msg, final Chann
     } catch (final Throwable throwable) {
       Tags.ERROR.set(span, Boolean.TRUE);
       span.log(Collections.singletonMap(ERROR_OBJECT, throwable));
-      Tags.HTTP_STATUS.set(span, 500);
+      NettyUtils.setServerSpanHttpStatus(span, 500);
```
keitwb: This is still assuming a 500 when we don't know for sure that the server is actually returning that to the client -- Or do we?

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
rmfitzpatrick: updated

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/server/HttpServerResponseTracingHandler.java
rmfitzpatrick: Updated

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/server/HttpServerResponseTracingHandler.java
```diff
@@ -28,12 +30,17 @@ public void write(final ChannelHandlerContext ctx, final Object msg, final Chann
     } catch (final Throwable throwable) {
       Tags.ERROR.set(span, Boolean.TRUE);
       span.log(Collections.singletonMap(ERROR_OBJECT, throwable));
-      Tags.HTTP_STATUS.set(span, 500);
+      try {
+        int status = response.getStatus().code();
+        NettyUtils.setServerSpanHttpStatus(span, status);
+      } catch (final Throwable exc) {
```
mpetazzoni: Why would this happen?

PR #22, opened 2019-04-02
Add Netty relocated package gradle property

rmfitzpatrick:
Will allow shade/shadow users to build an agent with nonstandard Netty package path:

`./gradlew build -PrelocatedNettyPackage='my.shaded.package.path.netty'`


tylerbenson:
I'm curious... How often do you see this come up?  Anything in particular?

rmfitzpatrick:
This would be for core frameworks that relocate their embedded server dependency, but could definitely extend to other use cases.

tylerbenson:
any specific examples?

PR #23, opened 2019-04-02
Add @TraceSetting and whitelist jax-rs-controller exceptions

rmfitzpatrick:
These changes introduce the `@TraceSetting` annotation for lower-level instrumentation configuration.  For the first provided feature, an `allowedExceptions` exception class array element will be used in determining whether to mark `jax-rs-controller` spans has having errored.

```
import com.signalfx.trace.api.TraceSetting;

@GET
@TraceSetting(allowedExceptions = {MyCustomException.class}
public void handle() throws MyCustomException {
  // This JAX resource will not be marked with error tags
  throw new MyCustomException();
}
```

### DIFF
--- a/dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
+++ b/dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
@@ -9,6 +9,7 @@ import static net.bytebuddy.matcher.ElementMatchers.isAnnotatedWith;
 import static net.bytebuddy.matcher.ElementMatchers.named;
 
 import com.google.auto.service.AutoService;
+import com.signalfx.tracing.api.TraceSetting;
 import datadog.trace.agent.tooling.Instrumenter;
 import io.opentracing.Scope;
 import io.opentracing.Span;
@@ -18,6 +19,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.Collections;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 import java.util.Stack;
 import javax.ws.rs.HttpMethod;
@@ -155,11 +157,25 @@ public final class JaxRsAnnotationsInstrumentation extends Instrumenter.Default 
 
     @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
     public static void stopSpan(
-        @Advice.Enter final Stack<Scope> scopes, @Advice.Thrown final Throwable throwable) {
+        @Advice.Origin final Method method,
+        @Advice.Enter final Stack<Scope> scopes,
+        @Advice.Thrown final Throwable throwable) {
       if (throwable != null) {
-        final Span span = scopes.peek().span();
-        Tags.ERROR.set(span, true);
-        span.log(Collections.singletonMap(ERROR_OBJECT, throwable));
+        List<Class> allowedExceptions = TraceSetting.Annotated.getAllowedExceptions(method);
+
+        Boolean setErrorTag = true;
+        for (Class allowed : allowedExceptions) {
+          if (allowed.isInstance(throwable)) {
+            setErrorTag = false;
+            break;
+          }
+        }
+
+        if (setErrorTag) {
+          final Span span = scopes.peek().span();
+          Tags.ERROR.set(span, true);
+          span.log(Collections.singletonMap(ERROR_OBJECT, throwable));
+        }
       }
 
       while (!scopes.isEmpty()) {
--- a/dd-trace-api/src/main/java/com/signalfx/tracing/api/TraceSetting.java
+++ b/dd-trace-api/src/main/java/com/signalfx/tracing/api/TraceSetting.java
@@ -1,0 +1,69 @@
+package com.signalfx.tracing.api;
+
+import static java.lang.annotation.ElementType.METHOD;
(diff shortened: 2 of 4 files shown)


dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
```diff
@@ -155,11 +153,38 @@ public JaxRsAnnotationsInstrumentation() {
...
+        @Advice.Enter final Stack<Scope> scopes,
+        @Advice.Thrown final Throwable throwable) {
       if (throwable != null) {
-        final Span span = scopes.peek().span();
-        Tags.ERROR.set(span, true);
-        span.log(Collections.singletonMap(ERROR_OBJECT, throwable));
+        ArrayList<Class> allowedExceptions = new ArrayList<>();
+
+        Class<?> target = method.getDeclaringClass();
+        final TraceSetting classAnnotation = target.getAnnotation(TraceSetting.class);
```
keitwb: This code duplication could be reduced by taking advantage of the fact that `Class` and `Method` both implement [AnnotatedElement](https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/AnnotatedElement.html).  In fact, if you can move this logic to the annotation class that would be even better.

dd-trace-api/src/main/java/com/signalfx/tracing/api/TraceSetting.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Use this annotation for class and method per-instrumentation configurations
+ *
+ */
+@Retention(RUNTIME)
+@Target({TYPE, METHOD})
+public @interface TraceSetting {
+
+  /** An array of Exceptions to ignore in jax-rs-controller auto-instrumented spans. */
```
keitwb: Comment says "jax-rs-controller" but this is in a generic package.  Think carefully about whether this is generic enough to share across multiple instrumentations.  Might be better to make this a jax-rs-controller specific annotation until more consideration is put into a generic one.  The fact that you commented it the way you did makes me think you mean for it to be jax-rs-controller specific for now.

rmfitzpatrick: I have thought about this being adopted by other instrumentations, as this use particular whitelist case can be extended for on method exit throwable checks project wide.  The comment is to denote that so far only the jax-rs annotation instrumentation has adopted it, and would be updated as that changes.

The trace setting annotation overall is designed to be a generic runtime configuration helper as providing one per-instrumentation doesn't scale.

dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
rmfitzpatrick: Good idea, updated.

dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
```diff
@@ -9,17 +9,15 @@
...
 import io.opentracing.Span;
 import io.opentracing.tag.Tags;
 import io.opentracing.util.GlobalTracer;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.Map;
-import java.util.Stack;
+import java.util.*;
```
mpetazzoni: Explicit imports are usually preferred

dd-trace-api/src/main/java/com/signalfx/tracing/api/TraceSetting.java
```diff
@@ -0,0 +1,69 @@
...
+import java.lang.annotation.Target;
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/** Use this annotation for class and method per-instrumentation configurations */
+@Retention(RUNTIME)
+@Target({TYPE, METHOD})
```
mpetazzoni: I find it a bit strange to have this on `TYPE` (for classes). Having a hard time picturing a use-case for setting this at the class level and how that relates to transaction flow for tracing.

rmfitzpatrick: For the allowed exceptions case, if you have a resource representing an endpoint with several http method processors, I think it would be preferable to annotate the class once instead of each method.  If some base resource is subclassed several times with the same desired whitelist, having to annotate each subclass would want to be avoided.

In general, I can see `TraceSetting` eventually being used to prevent specific classes from being instrumented without having to disable the instrumentation globally, so being able to annotate classes would be necessary for this case.

dd-java-agent/instrumentation/jax-rs-annotations/src/main/java/datadog/trace/instrumentation/jaxrs/JaxRsAnnotationsInstrumentation.java
rmfitzpatrick: Updated my ide setting.

rmfitzpatrick:
I'll provide proper documentation once this is adopted by the trace annotation.

PR #24, opened 2019-04-09
Initial RestTemplate instrumentation

rmfitzpatrick:
Adds RestTemplate tracing interceptor and sources it in updated spring-web tests.  AsyncRestTemplate is yet to be supported.


mpetazzoni:
Are the tests expected to pass though?

keitwb:
We need a bigger box on CircleCI for them to pass without running out of memory in Gradle.

PR #27, opened 2019-04-23
Provide informative Zipkin span names for web servers

rmfitzpatrick:
Web servers currently have extremely limited operation names that cause undesired grouping of dissimilar traces.  These changes take advantage of the normalizing and 404 path-masking resource name decorators for http.url tag values and update the Zipkin span names to be more specific.

If some service owners are using non-standard paths that don't match [the mixed alphanumeric pattern](https://github.com/signalfx/signalfx-java-tracing/blob/master/dd-trace-ot/src/main/java/datadog/opentracing/decorators/URLAsResourceName.java#L15), replacing operation name in gateway could be necessary: https://docs.signalfx.com/en/latest/integrations/integrations-reference/integrations.signalfx.gateway.html#identify-and-replace-variables-in-span-names.

### DIFF
--- a/dd-java-agent/instrumentation/jetty-8/src/main/java/datadog/trace/instrumentation/jetty8/JettyHandlerAdvice.java
+++ b/dd-java-agent/instrumentation/jetty-8/src/main/java/datadog/trace/instrumentation/jetty8/JettyHandlerAdvice.java
@@ -49,8 +49,8 @@ public class JettyHandlerAdvice {
     final Span span = scope.span();
     Tags.COMPONENT.set(span, "jetty-handler");
     Tags.HTTP_METHOD.set(span, req.getMethod());
-    Tags.HTTP_URL.set(span, req.getRequestURL().toString());
     span.setTag(DDTags.RESOURCE_NAME, resourceName);
+    Tags.HTTP_URL.set(span, req.getRequestURL().toString());
     if (req.getUserPrincipal() != null) {
       span.setTag(DDTags.USER_NAME, req.getUserPrincipal().getName());
     }
--- a/dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
+++ b/dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
@@ -2,6 +2,7 @@ package datadog.trace.common.writer;
 
 import static io.opentracing.tag.Tags.SPAN_KIND;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
@@ -131,6 +132,8 @@ public class ZipkinV2Api implements Api {
       tagNode.put(DDTags.RESOURCE_NAME, span.getResourceName());
     }
 
+    updateFromResourceTag(spanNode, tagNode);
+
     return spanNode;
   }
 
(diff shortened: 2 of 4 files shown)


dd-java-agent/instrumentation/jetty-8/src/main/java/datadog/trace/instrumentation/jetty8/JettyHandlerAdvice.java
```diff
@@ -50,7 +50,6 @@ public static Scope startSpan(
     Tags.COMPONENT.set(span, "jetty-handler");
     Tags.HTTP_METHOD.set(span, req.getMethod());
     Tags.HTTP_URL.set(span, req.getRequestURL().toString());
-    span.setTag(DDTags.RESOURCE_NAME, resourceName);
```
rmfitzpatrick: Removing this and setting a sane operation name default is necessary at the moment to prevent the ResourceNameDecorator from applying the instrumentation-set value over that of the URLAsResourceName in some cases, as their application order isn't guaranteed: https://github.com/signalfx/signalfx-java-tracing/blob/master/dd-trace-ot/src/main/java/datadog/opentracing/DDTracer.java#L81

dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
```diff
@@ -155,6 +157,23 @@ private String deriveKind(DDSpan span) {
     return null;
   }
 
+  /**
+   * In order to have more informative operation names for web frameworks, we take advantage of the
+   * URLAsResourceName-normalized resource name tag value, and update the operation name.
+   */
+  private void updateFromResourceTag(ObjectNode spanNode, ObjectNode tagNode) {
+    final String resourceName = tagNode.get(DDTags.RESOURCE_NAME).textValue();
```
keitwb: Is the null-safe?

mpetazzoni:
Seems weird to be doing this in the Zipkin-specific writer; I feel like this should be resolved before it reaches that layer.

rmfitzpatrick:
> Seems weird to be doing this in the Zipkin-specific writer; I feel like this should be resolved before it reaches that layer.

I agree but for the time being accomplishing this within the Zipkin translation layer allows us to more easily track w/ upstream and not have to make changes in all applicable instrumentations and their tests for the time being.  There will definitely be a point where we should shift this strategy more closely with [our initial version](https://github.com/signalfx/signalfx-java-tracing/blob/master-old/dd-java-agent/instrumentation/servlet-3/src/main/java/datadog/trace/instrumentation/servlet3/Servlet3Advice.java#L47).

dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
rmfitzpatrick: Updated.

dd-java-agent/instrumentation/jetty-8/src/main/java/datadog/trace/instrumentation/jetty8/JettyHandlerAdvice.java
```diff
@@ -49,8 +49,8 @@ public static Scope startSpan(
     final Span span = scope.span();
     Tags.COMPONENT.set(span, "jetty-handler");
     Tags.HTTP_METHOD.set(span, req.getMethod());
-    Tags.HTTP_URL.set(span, req.getRequestURL().toString());
     span.setTag(DDTags.RESOURCE_NAME, resourceName);
+    Tags.HTTP_URL.set(span, req.getRequestURL().toString());
```
rmfitzpatrick: Url needed to be set after resource name tag to prevent overwriting the decorator's change.

PR #31, opened 2019-05-14
Add initial Jersey tracing event listener

rmfitzpatrick:
Adds expanded Jersey instrumentation via an [`ApplicationEventListener`](https://jersey.github.io/apidocs/2.28/jersey/org/glassfish/jersey/server/monitoring/ApplicationEventListener.html).  This is helpful for cases where servers aren't instrumented as the JAX-RS annotation instrumentations are incapable of robustly extracting context from headers and getting response status codes.


dd-java-agent/instrumentation/jersey/src/main/java/datadog/trace/instrumentation/jersey/JerseyRequestExtractAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @Override
+  public Iterator<Map.Entry<String, String>> iterator() {
+    HashMap<String, String> delinkedListHeaders = new HashMap<>();
+
+    // Jersey headers are Map<String, List<String>> so multiple values need to be flattened
+    for (final Object e : headers.entrySet()) {
+      Map.Entry<String, List<String>> entry = (Map.Entry<String, List<String>>) e;
+      final String key = entry.getKey().toLowerCase();
+      final String val = joinCsv(entry.getValue());
```
mpetazzoni: I'm not sure joining multiple B3 propagation fields with commas is valid. Should this be looking for specific headers instead of flattening all?

rmfitzpatrick: This is just to get the headers in accepted carrier format for the tracer's actual extract method: https://github.com/signalfx/signalfx-java-tracing/blob/master/dd-trace-ot/src/main/java/datadog/opentracing/propagation/B3HttpCodec.java#L71

This is similar to other instrumentations like servlet: https://github.com/signalfx/signalfx-java-tracing/blob/master/dd-java-agent/instrumentation/servlet-3/src/main/java/datadog/trace/instrumentation/servlet3/HttpServletRequestExtractAdapter.java#L75.

Given that a number of frameworks allow multiple header values, it's definitely important for us to ensure that during injection we use whatever single value setter is available to never allow joined B3 values: https://github.com/signalfx/signalfx-java-tracing/commit/feee7369450164a46dd30af726f5e106678891fe.

rmfitzpatrick: On second thought it may make more sense just the use the first value instead of unnecessarily joining.  Nothing but the extraction is using this and it doesn't care about other headers.

dd-java-agent/instrumentation/jersey/src/main/java/datadog/trace/instrumentation/jersey/JerseyRequestExtractAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+    // Jersey headers are Map<String, List<String>> but B3 only supports single value
+    for (final Object e : headers.entrySet()) {
+      Map.Entry<String, List<String>> entry = (Map.Entry<String, List<String>>) e;
+      final String key = entry.getKey().toLowerCase();
+      // first item will do.
+      final String val = entry.getValue().get(0);
+      delinkedListHeaders.put(key, val);
+    }
+
+    return delinkedListHeaders.entrySet().iterator();
```
mpetazzoni: Another option altogether is to wrap the `headers.entrySet().iterator()` instead of building a new map:

```java
return new Iterator<Map.Entry<String, String>>() {
    private final Iterator<Map.Entry<String, List<String>>> iterator = headers.entrySet().iterator();
    public boolean hasNext() { return iterator.hasNext(); }
    public boolean remove() { return iterator.remove(); }
    public Map.Entry<String, String> next() {
      Map.Entry<String, List<String>> entry = iterator.next();
      return new AbstractMap.Entry<String, String>(entry.getKey(), entry.getValue().get(0));
    }
};
```

dd-java-agent/instrumentation/jersey/src/main/java/datadog/trace/instrumentation/jersey/JerseyRequestExtractAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+  JerseyRequestExtractAdapter(final ContainerRequest request) {
+    headers = request.getHeaders();
+  }
+
+  @Override
+  public Iterator<Map.Entry<String, String>> iterator() {
+    HashMap<String, String> delinkedListHeaders = new HashMap<>();
+
+    // Jersey headers are Map<String, List<String>> but B3 only supports single value
+    for (final Object e : headers.entrySet()) {
```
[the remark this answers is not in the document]
rmfitzpatrick: I was confused by this too.  Didn't realize I needed to be specific in the field declaration to resolve.

dd-java-agent/instrumentation/jersey/src/main/java/datadog/trace/instrumentation/jersey/JerseyRequestExtractAdapter.java
rmfitzpatrick: Updated w/ this approach, but there was some kind of helper class injector conflict when using an anonymous inner class.  I'll need to look into that later.

PR #32, opened 2019-05-21
Include tracer version in global tags

rmfitzpatrick:
Adds a library and version tag to the global defaults for reporting tracer information.

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/Config.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/Config.java
@@ -75,6 +75,15 @@ public class Config {
   public static final String LANGUAGE_TAG_VALUE = "jvm";
   public static final String DEFAULT_SERVICE_NAME = "unnamed-java-app";
 
+  public static final String TRACING_LIBRARY_KEY = "signalfx.tracing.library";
+  public static final String TRACING_LIBRARY_VALUE = "java-tracing";
+  public static final String TRACING_VERSION_KEY = "signalfx.tracing.version";
+  public static final String TRACING_VERSION_VALUE = "0.24.0-sfx9";
+  public static final String DEFAULT_GLOBAL_TAGS =
+      String.format(
+          "%s:%s,%s:%s",
+          TRACING_LIBRARY_KEY, TRACING_LIBRARY_VALUE, TRACING_VERSION_KEY, TRACING_VERSION_VALUE);
+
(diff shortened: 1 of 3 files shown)


dd-trace-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -155,7 +164,7 @@
         getBooleanSettingFromEnvironment(TRACE_RESOLVER_ENABLED, DEFAULT_TRACE_RESOLVER_ENABLED);
     serviceMapping = getMapSettingFromEnvironment(SERVICE_MAPPING, null);
 
-    globalTags = getMapSettingFromEnvironment(GLOBAL_TAGS, null);
+    globalTags = getMapSettingFromEnvironment(GLOBAL_TAGS, DEFAULT_GLOBAL_TAGS);
```
mpetazzoni: Does this still ad the `DEFAULT_GLOBAL_TAGS`, even if global tags are passed in from the environment/system properties?

dd-trace-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -75,6 +75,15 @@
   public static final String LANGUAGE_TAG_VALUE = "jvm";
   public static final String DEFAULT_SERVICE_NAME = "unnamed-java-app";
 
+  public static final String TRACING_LIBRARY_KEY = "signalfx.tracing.library";
+  public static final String TRACING_LIBRARY_VALUE = "java-tracing";
+  public static final String TRACING_VERSION_KEY = "signalfx.tracing.version";
+  public static final String TRACING_VERSION_VALUE = "0.24.0-sfx9";
```
mpetazzoni: We need to document our release process and make sure this gets updated properly when making a release.

dd-trace-api/src/main/java/datadog/trace/api/Config.java
rmfitzpatrick: No, but configuring the global tags wouldn't need to be advertised to users, who should be using `trace.span.tags` with the same functionality as we update the env var.

I could update `Config.getMergedSpanTags()` such that it's permanent if you'd prefer, though this was a bit less invasive and allows the option of overriding if desired.

mpetazzoni: That's fine

PR #33, opened 2019-05-31
Implement span.log()

rmfitzpatrick:
Currently `span.log()` is a noop unless a map with the OT error object field is provided, in which case the object is expanded as a set of error tags.  These changes introduce a general log mechanism and restrict the cases where the expanded error tagging is applied.  Not keeping the existing error tagging process warrants updating all instrumentation tests and is out of scope for this functionality.

### DIFF
--- a/dd-trace-ot/src/main/java/datadog/opentracing/DDSpan.java
+++ b/dd-trace-ot/src/main/java/datadog/opentracing/DDSpan.java
@@ -1,6 +1,10 @@
+// Modified by SignalFx
 package datadog.opentracing;
 
+import static io.opentracing.log.Fields.ERROR_KIND;
 import static io.opentracing.log.Fields.ERROR_OBJECT;
+import static io.opentracing.log.Fields.MESSAGE;
+import static io.opentracing.log.Fields.STACK;
 
 import com.fasterxml.jackson.annotation.JsonGetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -18,7 +22,10 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.lang.ref.WeakReference;
 import java.math.BigInteger;
+import java.util.AbstractMap;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
@@ -143,10 +150,14 @@ public class DDSpan implements Span, MutableSpan {
   }
 
   private boolean extractError(final Map<String, ?> map) {
-    if (map.get(ERROR_OBJECT) instanceof Throwable) {
-      final Throwable error = (Throwable) map.get(ERROR_OBJECT);
-      setErrorMeta(error);
-      return true;
+    final Object errorObject = map.get(ERROR_OBJECT);
+    if (errorObject instanceof Throwable) {
+      // if custom instrumentation don't use setErrorMeta
+      if (!map.containsKey(ERROR_KIND) && !map.containsKey(MESSAGE) && !map.containsKey(STACK)) {
+        final Throwable error = (Throwable) errorObject;
+        setErrorMeta(error);
+        return true;
+      }
     }
     return false;
   }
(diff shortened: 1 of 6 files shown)


dd-trace-ot/src/main/java/datadog/opentracing/DDSpan.java
```diff
@@ -144,9 +150,12 @@ public void setErrorMeta(final Throwable error) {
 
   private boolean extractError(final Map<String, ?> map) {
     if (map.get(ERROR_OBJECT) instanceof Throwable) {
-      final Throwable error = (Throwable) map.get(ERROR_OBJECT);
-      setErrorMeta(error);
-      return true;
+      // if custom instrumentation don't use setErrorMeta
+      if (map.get(ERROR_KIND) == null && map.get(MESSAGE) == null && map.get(STACK) == null) {
```
mpetazzoni: Use `map.containsKey()`

dd-trace-ot/src/main/java/datadog/opentracing/DDSpan.java
```diff
@@ -144,9 +150,12 @@ public void setErrorMeta(final Throwable error) {
...
+        final Throwable error = (Throwable) map.get(ERROR_OBJECT);
```
mpetazzoni: Get once to evaluate the `if` and cast as `Throwable` inside the if:

```java
Object obj = map.get(ERROR_OBJECT);
if (obj instanceof Throwable) {
  Throwable error = (Throwable) obj;
  setErrorMeta(error);
  return true;
}
```

Map lookups are expensive, and instrumentation must remain as cheap as possible.

dd-trace-ot/src/main/java/datadog/opentracing/DDSpanContext.java
```diff
@@ -321,6 +326,25 @@ public synchronized void setTag(final String tag, final Object value) {
     return Collections.unmodifiableMap(tags);
   }
 
+  /**
+   * Add a logged event map to the span. Logs are not propagated to the children.
+   *
+   * @param l the timestamp of the event.
```
mpetazzoni: Why not call the parameter `timestamp` then?

dd-trace-ot/src/main/java/datadog/opentracing/DDSpanContext.java
```diff
@@ -321,6 +326,25 @@ public synchronized void setTag(final String tag, final Object value) {
...
+   * @param map the event map for the tag.
+   */
+  public synchronized void log(final long l, final Map<String, ?> map) {
+    List<Map<String, ?>> logItems = logs.get(l);
+    if (logItems == null) {
```
mpetazzoni: Since we're in Java 8, you can replace the body of this method with:

```java
logs.computeIfAbsent(timestamp, ArrayList::new).add(map);
```

dd-trace-ot/src/main/java/datadog/opentracing/DDSpanContext.java
```diff
@@ -45,6 +47,9 @@
   /** Tags are associated to the current span, they will not propagate to the children span */
   private final Map<String, Object> tags = new ConcurrentHashMap<>();
 
+  /** Logs are associated to the current span, they will not propagate to the children span */
+  private final Map<Long, List<Map<String, ?>>> logs = new ConcurrentHashMap<>();
```
mpetazzoni: Does this really need to be a map indexed by timestamp? It could simply be a `List<Pair<Long, Map<String, ?>>>`

dd-trace-ot/src/main/java/datadog/opentracing/DDSpanContext.java
rmfitzpatrick: I think the source compatibility for the project is 7: https://github.com/signalfx/signalfx-java-tracing/blob/master/gradle/java.gradle#L14

dd-trace-ot/src/main/java/datadog/opentracing/DDSpanContext.java
rmfitzpatrick: I'd prefer not to introduce a new dependency (the project isn't currently using apache commons), so used `SimpleEntry` instead.  The list doesn't need to be instantiated via `synchronizedList()`, right?

dd-trace-ot/src/main/java/datadog/opentracing/DDSpanContext.java
mpetazzoni: Sigh, ok. Java 7 is EOL already, but not worth the effort here

dd-trace-ot/src/main/java/datadog/opentracing/DDSpanContext.java
mpetazzoni: If DDSpanContext isn't expected to be thread-safe, then no.

PR #34, opened 2019-06-04
Include upstream 0.28.0 additions

rmfitzpatrick:
These changes pull in 0.28.0 and provide necessary updates to the zipkin writer and B3 codec to reconcile dd's writer updates and B3 implementation.


mpetazzoni:
Not really reviewable but if the tests pass and it still works sending data to SignalFx; just :shipit:

PR #37, opened 2019-06-11
Update resource name

achandras:
Not sure this catches every case of the method name being used in the resource name, but at least it gets everything affected by the `URLAsResourceName` decorator.


rmfitzpatrick:
This is great.  Only things I see are w/ `latestDepTest`: https://github.com/signalfx/signalfx-java-tracing/blob/update_resource_name/dd-java-agent/instrumentation/ratpack-1.4/src/latestDepTest/groovy/RatpackTest.groovy and somehow netty spans are using the original value in SpringWebfluxTest: https://github.com/signalfx/signalfx-java-tracing/blob/update_resource_name/dd-java-agent/instrumentation/spring-webflux/src/test/groovy/SpringWebfluxTest.groovy

PR #38, opened 2019-06-12
Don't set error tag for 3xx or 4xx jax-rs client errors

rmfitzpatrick:
Since 3xx and 4xx status codes aren't considered to warrant error tags in traced clients, jax-rs exceptions thrown upon these responses should be ignored in instrumentation.  These changes whitelist `javax.ws.rs.core.Response`'s `RedirectionException` and `ClientErrorException`.

### DIFF
--- a/dd-java-agent/instrumentation/jax-rs-client/connection-error-handling-jersey/src/main/java/datadog/trace/instrumentation/connection_error/jersey/JerseyClientConnectionErrorInstrumentation.java
+++ b/dd-java-agent/instrumentation/jax-rs-client/connection-error-handling-jersey/src/main/java/datadog/trace/instrumentation/connection_error/jersey/JerseyClientConnectionErrorInstrumentation.java
@@ -1,3 +1,4 @@
+// Modified by SignalFx
 package datadog.trace.instrumentation.connection_error.jersey;
 
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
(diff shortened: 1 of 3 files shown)


dd-java-agent/instrumentation/jax-rs-client/connection-error-handling-jersey/src/main/java/datadog/trace/instrumentation/connection_error/jersey/JerseyClientConnectionErrorInstrumentation.java
```diff
@@ -62,12 +65,14 @@ public static void handleError(
         @Advice.FieldValue("requestContext") final ClientRequest context,
         @Advice.Thrown final Throwable throwable) {
       if (throwable != null) {
-
         final Object prop = context.getProperty(ClientTracingFilter.SPAN_PROPERTY_NAME);
         if (prop instanceof Span) {
           final Span span = (Span) prop;
-          Tags.ERROR.set(span, true);
-          span.log(Collections.singletonMap(Fields.ERROR_OBJECT, throwable));
+          if (!(throwable instanceof RedirectionException)
```
achandras: Is there a reason to prefer repeating this code rather than having it as a utility function somewhere?

rmfitzpatrick: Not sure the logic belongs in the `ClientTracingFilter` and introducing a new dependency that would be immediately reverted seems like wasted effort.  Ok with this landing and then adding an error logger to the decorators during 0.28.0 adoption?

PR #47, opened 2019-07-10
Use db.statement tag over resource.name for database decorators.

sm7x:


dd-java-agent/instrumentation/hibernate/core-3.3/src/test/groovy/QueryTest.groovy
```diff
@@ -105,24 +108,24 @@ class QueryTest extends AbstractHibernateTest {
...
+    "query.uniqueResult"  | "Value"                           | "from Value where id = 1"           | false               | { sess ->
       Query q = sess.createQuery("from Value where id = 1")
       q.uniqueResult()
     }
-    "iterate"             | "from Value"                     | false               | { sess ->
-      Query q = sess.createQuery("from Value")
-      q.iterate()
-    }
-    "query.scroll"        | "from Value"                     | false               | { sess ->
+    // "iterate"             | "from Value"                      | "from Value"                     | false               | { sess ->
```
rmfitzpatrick: rather than commenting out, please address why the test isn't working as desired.  Same with the span decorator test.

dd-java-agent/instrumentation/hibernate/core-4.0/src/test/groovy/QueryTest.groovy
```diff
@@ -177,6 +180,7 @@ class QueryTest extends AbstractHibernateTest {
           tags {
             "$Tags.COMPONENT.key" "java-hibernate"
             "$Tags.SPAN_KIND.key" Tags.SPAN_KIND_CLIENT
+            // "$Tags.DB_STATEMENT.key" "from Value"
```
rmfitzpatrick: oversight or not working?

dd-java-agent/instrumentation/jedis-1.4/src/test/groovy/JedisClientTest.groovy
```diff
@@ -84,10 +86,10 @@ class JedisClientTest extends AgentTestRunner {
       trace(1, 1) {
         span(0) {
           serviceName "unnamed-java-app"
-          operationName "redis.query"
-          resourceName "GET"
+          operationName "redis.query"
```
rmfitzpatrick: is the resource name no longer `GET`?

dd-java-agent/instrumentation/jedis-1.4/src/test/groovy/JedisClientTest.groovy
```diff
@@ -112,8 +114,10 @@ class JedisClientTest extends AgentTestRunner {
           serviceName "unnamed-java-app"
           operationName "redis.query"
           resourceName "SET"
+
```
rmfitzpatrick: nit

dd-java-agent/instrumentation/mongo-3.1/src/main/java/datadog/trace/instrumentation/mongo/MongoClientDecorator.java
```diff
@@ -70,6 +71,7 @@ public Span onStatement(final Span span, final BsonDocument statement) {
     final String mongoCmd = scrubbed.toString();
 
     span.setTag(DDTags.RESOURCE_NAME, mongoCmd);
+    Tags.DB_STATEMENT.set(span, mongoCmd);
```
rmfitzpatrick: Is this not already done by the subsequent `DatabaseClientDecorator` `onStatement`?

rmfitzpatrick:
Can you add truncation to https://github.com/signalfx/signalfx-java-tracing/blob/master/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/decorator/DatabaseClientDecorator.java#L39?

dd-java-agent/instrumentation/hibernate/core-3.3/src/test/groovy/SessionTest.groovy
```diff
@@ -473,11 +475,11 @@ class SessionTest extends AbstractHibernateTest {
       }
     }
 
-    where:
-    queryMethodName  | resource              | queryBuildMethod
-    "createQuery"    | "Value"               | { sess -> sess.createQuery("from Value") }
-    "getNamedQuery"  | "Value"               | { sess -> sess.getNamedQuery("TestNamedQuery") }
-    "createSQLQuery" | "SELECT * FROM Value" | { sess -> sess.createSQLQuery("SELECT * FROM Value") }
+     where:
```
rmfitzpatrick: nit

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -45,7 +45,14 @@ protected String dbUser(final JDBCMaps.DBInfo info) {
 
   @Override
   protected String dbInstance(final JDBCMaps.DBInfo info) {
-    return info.getUrl();
+    String dbName = null;
+    String urlInfo = info.getUrl();
+    String[] urlArray = urlInfo.split("/");
+    if (urlArray.length > 1) {
```
rmfitzpatrick: Is this condition necessary?

dd-java-agent/instrumentation/hibernate/core-3.3/src/test/groovy/QueryTest.groovy
sm7x: I am not sure why they are not working as desired, left them there so I could get insight from you.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
sm7x: > Can you add truncation to https://github.com/signalfx/signalfx-java-tracing/blob/master/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/decorator/DatabaseClientDecorator.java#L39?

You want me to truncate the `DB_STATEMENT`? I thought we always allowed the full statement?

dd-java-agent/instrumentation/hibernate/core-4.0/src/test/groovy/QueryTest.groovy
sm7x: Thought I took that out. But that test doesn't work, same as the `query.iterate` test above.

dd-java-agent/instrumentation/jedis-1.4/src/test/groovy/JedisClientTest.groovy
sm7x: It is. I initially took out all `resourceName`s, since I wasn't sure we should be testing spans for them once they are filtered out. I omitted putting  this back.

dd-java-agent/instrumentation/hibernate/core-3.3/src/test/groovy/QueryTest.groovy
sm7x: For the span decorator test, we already took out what was preventing the DBStatement from interacting with Mongo queries (https://github.com/signalfx/signalfx-java-tracing/blob/904e4cbb8c74e0a0b379be27ba7bf20c9a3f774b/dd-trace-ot/src/main/java/datadog/opentracing/decorators/DBStatementAsResourceName.java#L26) 

So testing that will fail.

rmfitzpatrick: Nothing involving the earlier mongo component bypass should have changed.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
sm7x: > Is this condition necessary?

If `urlInfo` is empty or contains no slashes, then I think it will be helpful to have that check.

dd-java-agent/instrumentation/hibernate/core-4.0/src/test/groovy/QueryTest.groovy
sm7x: Took it out. `Iterate` is rather expensive as it could end up making several calls to the database for each record. I believe this may get pretty chatty/expensive... (`List` on the other hand just gets all the info in one call. Think `scroll` is similar to list but kinda truncated... I haven't fully confirmed this)

sm7x: `iterate` is still tested separately (below the first test testing a list of functions)

dd-trace-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -126,6 +126,9 @@
 
   public static final int DEFAULT_JMX_FETCH_STATSD_PORT = 8125;
 
+  public static final int MAX_DB_STATEMENT_VALUE_SIZE = 1024;
+  public static final String DB_STATEMENT_VALUE_SIZE = "db.statement.value.size";
```
rmfitzpatrick: maybe `db.statement.max.length`?  Also, please add to an existing config test

dd-java-agent/instrumentation/mongo-3.1/src/test/groovy/MongoClientDecoratorTest.groovy
```diff
@@ -34,9 +35,8 @@ class MongoClientDecoratorTest extends Specification {
     DECORATE.onStatement(span, query)
 
     then:
-    1 * span.setTag(Tags.DB_STATEMENT.key, expected)
     1 * span.setTag(DDTags.RESOURCE_NAME, expected)
-    0 * _
+    _ * Tags.DB_STATEMENT.set(span, expected)
```
rmfitzpatrick: Why should any of these need to change?

dd-trace-ot/src/main/java/datadog/opentracing/decorators/DBStatementAsResourceName.java
```diff
@@ -23,15 +24,9 @@ public boolean shouldSetTag(final DDSpanContext context, final String tag, final
...
-      // TODO: remove properly the tag (immutable at this time)
-      // the `db.statement` tag must be removed because it will be set
-      // by the Datadog Trace Agent as `sql.query`; here we're removing
-      // a duplicate that will not be obfuscated with the current Datadog
-      // Trace Agent version.
-      context.setTag(Tags.DB_STATEMENT.getKey(), null);
-      return false;
-    }
+    // if (!super.shouldSetTag(context, tag, value)) {
+    //   return true;
```
rmfitzpatrick: Instead of commenting out sections of code, would it make more sense to write the intended behavior and then provide a rationale?

dd-trace-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -126,6 +126,9 @@
 
   public static final int DEFAULT_JMX_FETCH_STATSD_PORT = 8125;
 
+  public static final int DEFAULT_DB_STATEMENT_MAX_LENGTH = 1024;
+  public static final String DB_STATEMENT_MAX_LENGTH = "db.statement.max.length";
```
rmfitzpatrick: Do you mind moving this up with the other system properties and documenting it in the readme?

dd-trace-api/src/test/groovy/datadog/trace/api/ConfigTest.groovy
```diff
@@ -292,6 +302,7 @@ class ConfigTest extends Specification {
     environmentVariables.set(DD_SERVICE_NAME_ENV, "still something else")
     environmentVariables.set(DD_WRITER_TYPE_ENV, "LoggingWriter")
     environmentVariables.set(DD_TRACE_AGENT_PORT_ENV, "777")
+    environmentVariables.set(SIGNALFX_DB_STATEMENT_MAX_LENGTH, "105")
```
rmfitzpatrick: I think the point of this suite is to ensure system properties take precedence over env vars if both set.  Mind adding a different property value and assertion like the other config options?

sm7x: Oh, I actually removed that. I am not sure, do we want system properties to over ride the env vars for the db_statement_length? Since the env var (if available) will be specified by the user?

PR #50, opened 2019-07-12
Update and document span tags configuration

rmfitzpatrick:
Updates the span tags property/env var info and adds configuration information to readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,50 +1,51 @@
 # SignalFx Java Agent
 
-This is the SignalFx Java Agent, a JVM agent to automatically instrument your
-Java application to capture and report distributed traces to SignalFx.
+This is the SignalFx Java Agent, a [JVM agent](https://docs.oracle.com/javase/7/docs/api/java/lang/instrument/package-summary.html)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -73,9 +75,54 @@ given in the table below.
...
+  $ export SIGNALFX_SERVICE_NAME=my-app
+  $ export SIGNALFX_SPAN_TAGS="MyTag:MyValue"
+  $ java -javaagent:signalfx-tracing.jar <rest of command-line arguments...>
+```
+
+The standard invocation assumes that you have a [locally deployed SignalFx Smart Agent](https://docs.signalfx.com/en/latest/apm/apm-deployment/smart-agent.html)
+accessible on `localhost`.  If you want to send traces directly to the Smart Gateway or if your Smart Agent is only
+accessible on a different host (e.g. in a containerized environment), set the trace host via the `signalfx.agent.host`
+property to the appropriate value (default is `localhost`).  You can override the entire trace target URL with the
+`signalfx.endpoint.url` property.
```
mpetazzoni: Mention that this can also be done through the corresponding env variables.

PR #51, opened 2019-07-15
Use redis command as operation name

rmfitzpatrick:
Provides more informative operation names for redis instrumentation.

### DIFF
--- a/dd-java-agent/instrumentation/jedis-1.4/src/main/java/datadog/trace/instrumentation/jedis/JedisInstrumentation.java
+++ b/dd-java-agent/instrumentation/jedis-1.4/src/main/java/datadog/trace/instrumentation/jedis/JedisInstrumentation.java
@@ -1,3 +1,4 @@
+// Modified by SignalFx
 package datadog.trace.instrumentation.jedis;
 
 import static datadog.trace.instrumentation.jedis.JedisClientDecorator.DECORATE;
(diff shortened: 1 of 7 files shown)


dd-java-agent/instrumentation/jedis-1.4/src/main/java/datadog/trace/instrumentation/jedis/JedisInstrumentation.java
```diff
@@ -59,6 +60,7 @@ public JedisInstrumentation() {
     @Advice.OnMethodEnter(suppress = Throwable.class)
     public static Scope startSpan(@Advice.Argument(1) final Command command) {
       final Scope scope = GlobalTracer.get().buildSpan("redis.command").startActive(true);
+      scope.span().setOperationName(command.name());
```
mpetazzoni: Why not do that in the `buildSpan()` above? Also, I think it would be more readable for the user if the operation name still said `redis`:

```java
final Scope scope = GlobalTracer.get()
    .buildSpan("redis." + command.name().toLowerString())
    .startActive(true);
```

dd-java-agent/instrumentation/lettuce-5/src/main/java8/datadog/trace/instrumentation/lettuce/LettuceClientDecorator.java
```diff
@@ -60,6 +61,7 @@ public Span onConnection(final Span span, final RedisURI connection) {
 
   public Span onCommand(final Span span, final RedisCommand command) {
     final String commandName = LettuceInstrumentationUtil.getCommandName(command);
+    span.setOperationName(commandName);
```
mpetazzoni: Same here:

```java
span.setOperationName("redis." + commandName.toLowerCase());
```

mpetazzoni:
@rmfitzpatrick Actually depending on how we want to best leverage the `component` tag in the future maybe we don't need to prefix with `redis.`. Thoughts?

rmfitzpatrick:
> @rmfitzpatrick Actually depending on how we want to best leverage the `component` tag in the future maybe we don't need to prefix with `redis.`. Thoughts?

I think leveraging existing spans and removing redundancy is a good practice overall, as this could be similar to our earlier framework operation names and the http method tag.

PR #54, opened 2019-07-17
Stop JDBC setting resource.name to db statement

sm7x:


dd-java-agent/instrumentation/hibernate/core-3.3/src/test/groovy/SessionTest.groovy
```diff
@@ -83,14 +85,14 @@ class SessionTest extends AbstractHibernateTest {
     }
 
     where:
-    testName  | methodName | resource | sessionImplementations                    | sessionMethodTest
-    "lock"    | "lock"     | "Value"  | [sessionBuilder]                          | { sesh, val ->
+    testName  | methodName | resource | sessionImplementations                      | sessionMethodTest
```
rmfitzpatrick: nit: please avoid all unnecessary whitespace changes unless they improve readability in documentation.  I don't think columns need to line up to be a valid data table.

sm7x: I found it easier to read/grasp with the columns aligned, that's why I did that.

PR #61, opened 2019-08-01
Document support of Hibernate in README

sm7x:
* Also update version support for Jedis and Vertx


rmfitzpatrick:
Looks like we support 3.3+ https://github.com/signalfx/signalfx-java-tracing/blob/master/dd-java-agent/instrumentation/hibernate/core-3.3/core-3.3.gradle#L35

sm7x:
> Looks like we support 3.3+ https://github.com/signalfx/signalfx-java-tracing/blob/master/dd-java-agent/instrumentation/hibernate/core-3.3/core-3.3.gradle#L35
> 
> @rmfitzpatrick Yes, it does, but Mohit suggested a conservative 3.5+ for the product docs. I believe both product docs and the README should be aligned.

PR #66, opened 2019-10-01
khttp instrumentation

rmfitzpatrick:
Adds instrumentation for khttp requests: https://khttp.readthedocs.io/en/latest/


dd-java-agent/instrumentation/khttp/src/main/java8/datadog/trace/instrumentation/khttp/KHttpAdvice.java
```diff
@@ -0,0 +1,80 @@
...
+      }
+    }
+
+    if (!awsClientCall) {
+      Class emptyMap = KHttpAdviceUtils.emptyMap;
+      if (emptyMap != null && emptyMap.isInstance(headers)) {
+        // EmptyMap is read-only so we need to overwrite default arg for header injection
+        headers = new HashMap<String, String>();
+      }
+      tracer.inject(
```
mpetazzoni: Should we offer a way to turn outbound propagation off? Not sure if we do that in `apache-httpclient`. If not, we should, but maybe tracked as a subsequent story.

rmfitzpatrick: I think ideally we'd provide some url/domain-based whitelist configuration mechanism to be used in all http clients.  Don't believe any of the other clients have a toggle property as is.

PR #69, opened 2019-11-07
Add 128bit support to B3

rmfitzpatrick:
Currently we drop the upper 64 bits for 128 bit trace ids, which is unnecessary and undesired.


rmfitzpatrick:
> Looks good.
> 
> So basically you just needed to change the validator to not reject between 64 and 128 bit ids but nothing else needs to change?

It appears so since the ids are stored as strings in span contexts.

PR #70, opened 2019-11-07
Add redis db.statement

rmfitzpatrick:
Adds redis command arguments with `db.statement` tag and provides a new `signalfx.instrumentation.redis.capture-command-arguments` property to allow disabling this feature.


rmfitzpatrick:
> It's unfortunate that the same code is repeated three times. Can this be refactored?

Are you referring to the two repeated for loops in the jedis instrumentations?  If I were to break out a `datadog.trace.agent.decorator.RedisClientDecorator` with methods to retrieve command arg strings I think I'd still have to repeat the implementation for each version?

rmfitzpatrick:
@mpetazzoni, I believe I've incorporated all of your feedback.

Issue #68, opened 2019-11-07
MDCInjectionInstrumentation displays trace_id and span_id as longs


billoneil:
This is more of a UX issue than a bug but the MDCInjectionInstrumentation logs `signalfx.trace_id` and `signalfx.span_id` as longs where every reference in the signalfx UI is the HEX format.

Would it make sense to change the format to be HEX to match the UI or add an additional Instrumentation object that uses HEX instead to not break backwards compatibility? Would be happy to contribute something here would just know which direction you would like to go.

rmfitzpatrick:
@billoneil, thank you for your suggestion and your patience!  We'll be shortly providing this change in the next release.

PR #75, opened 2020-02-05
Trace annotation and MDC usability improvements

rmfitzpatrick:
These changes ensure that non-DD trace annotations use the same behavior as that of `com.signalfx.tracing.api.Trace` regarding operation names, and also ensure that logged MDC entries use hex representation for active span and its trace ids (resolves https://github.com/signalfx/signalfx-java-tracing/issues/68).  The latter required relocating the ids helper to the api project to prevent an unwanted dependency on the agent.

Also bumping up test timeout to reduce flake.

### DIFF
--- a/dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/TraceAdvice.java
+++ b/dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/TraceAdvice.java
@@ -18,34 +18,30 @@ public class TraceAdvice {
 
   @Advice.OnMethodEnter(suppress = Throwable.class)
   public static AgentScope onEnter(@Advice.Origin final Method method) {
-    Annotation trace = method.getAnnotation(Trace.class);
-    if (trace == null) {
-      trace = method.getAnnotation(com.signalfx.tracing.api.Trace.class);
-    }
+    String resourceName = DECORATE.spanNameForMethod(method);
+    String operationName = resourceName;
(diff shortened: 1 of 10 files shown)


dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/TraceAdvice.java
```diff
@@ -18,24 +18,22 @@
...
   public static AgentScope onEnter(@Advice.Origin final Method method) {
-    Annotation trace = method.getAnnotation(Trace.class);
-    if (trace == null) {
-      trace = method.getAnnotation(com.signalfx.tracing.api.Trace.class);
-    }
-
-    boolean useResourceAsOperation = false;
     String operationName = null;
     String resourceName = null;
+    boolean useResourceAsOperation = true;
```
tigrannajaryan: The logic of when the resource is used as an operation name is not entirely clear. Is there a way to refactor this to make more understandable? I have a hard time following all possible code paths where useResourceAsOperation, operationName and resourceName are modified.

rmfitzpatrick: @tigrannajaryan is this better?  I removed the flag and opted for saner defaults.

tigrannajaryan: Much better, thank you.

PR #77, opened 2020-02-14
Include support for play framework 2.8

sm7x:


dd-java-agent/instrumentation/play-2.8/src/main/java8/datadog/trace/instrumentation/play28/PlayHttpServerDecorator.java
```diff
@@ -0,0 +1,96 @@
...
+        final String path = defOption.get().path();
+        span.setTag(DDTags.RESOURCE_NAME, path);
+      }
+    }
+    return span;
+  }
+
+  @Override
+  public AgentSpan onError(final AgentSpan span, Throwable throwable) {
+    span.setTag(Tags.HTTP_STATUS, 500);
```
rmfitzpatrick: It's advisable to only set the status if made available by a response object.  I'd imagine it's possible for this to be incorrect and could cause issues for users.

dd-java-agent/instrumentation/play-2.8/play-2.8.gradle
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+dependencies {
+  main_java8Compile group: 'com.typesafe.play', name: "play_$scalaVersion", version: playVersion
+
+  testCompile project(':dd-java-agent:instrumentation:java-concurrent')
+  testCompile project(':dd-java-agent:instrumentation:netty-4.0')
+  testCompile project(':dd-java-agent:instrumentation:netty-4.1')
+  testCompile project(':dd-java-agent:instrumentation:akka-http-10.0')
```
rmfitzpatrick: If you add the play 2.4 and 2.6 projects are there any conflicts or issues?

sm7x: Issues: All the tests for the 2.8 async server fail.

rmfitzpatrick: That is likely a sign this instrumentation, as is, introduces breaking behavior.

dd-java-agent/instrumentation/play-2.6/src/main/java8/datadog/trace/instrumentation/play26/PlayHttpServerDecorator.java
```diff
@@ -66,12 +68,27 @@ public AgentSpan onRequest(final AgentSpan span, final Request request) {
...
-        span.setTag(DDTags.RESOURCE_NAME, path);
+      Class HandlerDefClass = Router.Attrs.HANDLER_DEF.getClass();
+      Method asScalaOrUnderlying = null;
+
+      try {
+        asScalaOrUnderlying = HandlerDefClass.getMethod("underlying");
+      } catch (NoSuchMethodException e) {
+        try {
+          asScalaOrUnderlying = HandlerDefClass.getMethod("asScala");
+        } catch (NoSuchMethodException ee) {}
```
rmfitzpatrick: Mind logging this case?

dd-java-agent/instrumentation/play-2.6/src/main/java8/datadog/trace/instrumentation/play26/PlayHttpServerDecorator.java
```diff
@@ -66,12 +68,27 @@ public AgentSpan onRequest(final AgentSpan span, final Request request) {
...
+      try {
+        final Option<HandlerDef> defOption =
+          request.attrs().get((TypedKey) asScalaOrUnderlying.invoke(Router.Attrs.HANDLER_DEF));
+        if (!defOption.isEmpty()) {
+          final String path = defOption.get().path();
+          span.setTag(DDTags.RESOURCE_NAME, path);
+        }
       }
+      catch (IllegalAccessException e) {}
+      catch (InvocationTargetException e) {}
```
rmfitzpatrick: Same logging request

rmfitzpatrick: Ideally you'd only need to wrap the invoke call and not the path access and tagging.

sm7x: `defOption` would be out of scope if I wrap only the invoke call.

rmfitzpatrick: In that case you wouldn't mark it final and do a null check.

dd-java-agent/instrumentation/play-2.6/src/main/java8/datadog/trace/instrumentation/play26/PlayHttpServerDecorator.java
```diff
@@ -66,8 +68,20 @@ public AgentSpan onRequest(final AgentSpan span, final Request request) {
...
+      } catch (NoSuchMethodException e) {
+        try {
+          asScalaOrUnderlying = HandlerDefClass.getMethod("underlying");
+        } catch (NoSuchMethodException err) {
+          log.error(err.getMessage());
+        }
+      }
+
+      final Option<HandlerDef> defOption = getDefOption(request, asScalaOrUnderlying);
       if (!defOption.isEmpty()) {
```
rmfitzpatrick: You'll want to add a null check based on how `getDefOption` works.

dd-java-agent/instrumentation/play-2.6/src/main/java8/datadog/trace/instrumentation/play26/PlayHttpServerDecorator.java
```diff
@@ -92,4 +93,29 @@ public AgentSpan onError(final AgentSpan span, Throwable throwable) {
...
+    Class HandlerDefClass = handler.getClass();
+    Method asScalaOrUnderlying = null;
+
+    try {
+      asScalaOrUnderlying = HandlerDefClass.getMethod("asScala");
+    } catch (NoSuchMethodException e) {
+      try {
+        asScalaOrUnderlying = HandlerDefClass.getMethod("underlying");
+      } catch (NoSuchMethodException err) {
+        log.error(err.getMessage());
```
rmfitzpatrick: we should return here since the subsequent invoke will throw an NPE

dd-java-agent/instrumentation/play-2.6/src/main/java8/datadog/trace/instrumentation/play26/PlayHttpServerDecorator.java
```diff
@@ -92,4 +93,30 @@ public AgentSpan onError(final AgentSpan span, Throwable throwable) {
...
+        return null;
+      }
+    }
+
+    Option<HandlerDef> defOption = null;
+    try {
+      defOption = request.attrs().get((play.api.libs.typedmap.TypedKey) asScalaOrUnderlying.invoke(Router.Attrs.HANDLER_DEF));
```
rmfitzpatrick: It's preferable to check before casting:
```suggestion
      Object typedKey = null;
      try {
        typedKey = asScalaOrUnderlying.invoke(Router.Attrs.HANDLER_DEF);
      } catch (ReflectiveOperationException e) {
        log.error(e.getMessage());
      }
      if (typedKey instanceof play.api.libs.typedmap.TypedKey) {
        return request.attrs().get((play.api.libs.typedmap.TypedKey) typedKey);
      } else {
        log.error("Unexpected type for TypedMap key: " + typedKey);
      }
      return null;
```

I've also opted to use the base exception type.

PR #79, opened 2020-02-19
Add Zipkin recorded value max length config option

rmfitzpatrick:
These changes introduce the `recorded.value.max.length` config option to prevent Zipkin-encoded tag and log values from exceeding a specified size (default of 12288 chars).

### DIFF
--- a/dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
+++ b/dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.base.Strings;
 import datadog.opentracing.DDSpan;
+import datadog.trace.api.Config;
 import datadog.trace.api.DDSpanTypes;
 import datadog.trace.api.DDTags;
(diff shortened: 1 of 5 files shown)


dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
```diff
@@ -273,6 +276,14 @@ private static HttpURLConnection getHttpURLConnection(final String endpoint) thr
     return httpCon;
   }
 
+  private String truncatedString(String value) {
+    String encodedValue = value;
+    if (encodedValue.length() > recordedValueMaxLength) {
+      encodedValue = encodedValue.substring(0, recordedValueMaxLength) + "[...]";
```
tigrannajaryan: Is it ok that we append 5 characters so the value's length actually exceeds `recordedValueMaxLength`?

rmfitzpatrick: I wondered about this too.  It could be poor ux to not signal having truncated a value as well as to not write as many recorded characters as specified by the configuration.  Since the max length is relatively arbitrary I thought this approach was preferable to the alternatives.

tigrannajaryan: Sounds reasonable.

PR #81, opened 2020-03-05
Add Trace annotation blacklist configurable

rmfitzpatrick:
These changes add a new configuration property `signalfx.trace.annotated.method.blacklist` that will turn the existing `TraceAdvice` methods into a noop.  This approach was taken to avoid the complexity of converting the `TraceAnnotationInstrumentation` to something like the current `TraceConfigInstrumentation`.  It will also ensure that (in)directly annotated methods will not be traced across instrumentation approaches.*

Also cherry picking restricting latest Tomcat test dependencies.

### DIFF
--- a/dd-java-agent/instrumentation/jsp-2.3/jsp-2.3.gradle
+++ b/dd-java-agent/instrumentation/jsp-2.3/jsp-2.3.gradle
@@ -32,7 +32,7 @@ dependencies {
 
   latestDepTestCompile group: 'javax.servlet.jsp', name: 'javax.servlet.jsp-api', version: '+'
   latestDepTestCompile group: 'javax.servlet', name: 'javax.servlet-api', version: '+'
-  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-core', version: '+'
-  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-jasper', version: '+'
-  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-logging-juli', version: '+'
+  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-core', version: '9.+'
+  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-jasper', version: '9.+'
+  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-logging-juli', version: '9.+'
 }
--- a/dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/TraceAnnotationUtils.java
+++ b/dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/TraceAnnotationUtils.java
@@ -16,7 +16,8 @@ public class TraceAnnotationUtils {
       getClassMethodMap(Config.get().getAnnotatedMethodBlacklist());
 
   static final String PACKAGE_CLASS_NAME_REGEX = "[\\w.\\$]+";
-  private static final String METHOD_LIST_REGEX = "\\s*(?:\\w+\\s*,)*\\s*(?:\\w+\\s*,?)\\s*";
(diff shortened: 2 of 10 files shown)


dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/TraceAnnotationUtils.java
```diff
@@ -0,0 +1,65 @@
...
+      final Map<String, Set<String>> toTrace = new HashMap<>(0);
+      final String[] classMethods = configString.split(";", -1);
+      for (final String classMethod : classMethods) {
+        if (classMethod.trim().isEmpty()) {
+          continue;
+        }
+        final String[] splitClassMethod = classMethod.split("\\[", -1);
+        final String className = splitClassMethod[0];
+        final String method = splitClassMethod[1].trim();
+        final String methodNames = method.substring(0, method.length() - 1);
```
johnbley: A nice feature might be to support wildcarding with "someclass[*]".  This could be implemented by something like
if (methodName.equals("*")) {
  toTrace.put(className.trim(), new ContainsEverythingSet()); // or simple closure as such overriding contains().
} else { 
   // original logic here
But I'd leave that to your discretion based on, e.g., time pressure.

dd-java-agent/instrumentation/jsp-2.3/jsp-2.3.gradle
```diff
@@ -32,7 +32,7 @@ dependencies {
 
   latestDepTestCompile group: 'javax.servlet.jsp', name: 'javax.servlet.jsp-api', version: '+'
   latestDepTestCompile group: 'javax.servlet', name: 'javax.servlet-api', version: '+'
-  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-core', version: '+'
-  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-jasper', version: '+'
-  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-logging-juli', version: '+'
+  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-core', version: '9.+'
+  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-jasper', version: '9.+'
+  latestDepTestCompile group: 'org.apache.tomcat.embed', name: 'tomcat-embed-logging-juli', version: '9.+'
 }
```
johnbley: For the record, this cherry-picked patch made my tests pass as well, whereas previously they reliably failed.

dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/TraceAnnotationUtils.java
```diff
@@ -0,0 +1,65 @@
...
+          + PACKAGE_CLASS_NAME_REGEX
+          + "\\["
+          + METHOD_LIST_REGEX
+          + "\\]\\s*;)*\\s*"
+          + PACKAGE_CLASS_NAME_REGEX
+          + "\\["
+          + METHOD_LIST_REGEX
+          + "\\]\\s*;?\\s*";
+
+  public static Map<String, Set<String>> getClassMethodMap(final String configString) {
```
johnbley: Thanks for refactoring this out rather than copy/paste/tweaking it.

johnbley:
LGTM, +1, Fast Shipping, A++++++ would buy again

dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/TraceAnnotationUtils.java
rmfitzpatrick: Great idea, so I've added this.  Mind taking another look?

rmfitzpatrick:
> Nice name, I wonder how you chose it?

I may have been primed a bit, but it came to me effortlessly.

PR #82, opened 2020-03-13
Add maxSpansPerTrace config parameter, with a default of 2000.

johnbley:
This limits the writing of traces with more spans than the configured max.
Everything else about spans (propagation, id chains, etc.) still works;
only "at the last minute" do we drop the span before putting on the wire
to the upstream trace receiver.  2000 as the default was chosen somewhat
arbitrarily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ content](https://docs.signalfx.com/en/latest/apm/apm-overview/apm-metadata.html)
 | `signalfx.db.statement.max.length` | `SIGNALFX_DB_STATEMENT_MAX_LENGTH` | `1024` | The maximum number of characters written for the OpenTracing `db.statement` tag. |
 | `signalfx.recorded.value.max.length` | `SIGNALFX_RECORDED_VALUE_MAX_LENGTH` | `12288` | The maximum number of characters for any Zipkin-encoded tagged or logged value. |
 | `signalfx.trace.annotated.method.blacklist` | `SIGNALFX_TRACE_ANNOTATED_METHOD_BLACKLIST` | `null` | Prevent `@Trace` annotation functionality for the target method string of format `package.OuterClass[methodOne,methodTwo];other.package.OuterClass$InnerClass[*];` (`;` is required and `*` for all methods in class). |
+| `signalfx.max.spans.per.trace` | `SIGNALFX_MAX_SPANS_PER_TRACE` | `2000` | Drop (don't deliver) spans with more traces than this. Intended to prevent runaway traces from flooding upstream systems. |
 
 **Note: System property values take priority over corresponding environment
 variables.**
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ content](https://docs.signalfx.com/en/latest/apm/apm-overview/apm-metadata.html)
 | `signalfx.db.statement.max.length` | `SIGNALFX_DB_STATEMENT_MAX_LENGTH` | `1024` | The maximum number of characters written for the OpenTracing `db.statement` tag. |
 | `signalfx.recorded.value.max.length` | `SIGNALFX_RECORDED_VALUE_MAX_LENGTH` | `12288` | The maximum number of characters for any Zipkin-encoded tagged or logged value. |
 | `signalfx.trace.annotated.method.blacklist` | `SIGNALFX_TRACE_ANNOTATED_METHOD_BLACKLIST` | `null` | Prevent `@Trace` annotation functionality for the target method string of format `package.OuterClass[methodOne,methodTwo];other.package.OuterClass$InnerClass[*];` (`;` is required and `*` for all methods in class). |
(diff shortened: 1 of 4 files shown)


johnbley:
Comments particularly welcome on a better default value, hopefully derived in some fashion other than alchemical magic.

rmfitzpatrick:
In cases where the `PARTIAL_FLUSH_MIN_SPANS` functionality is triggered (scope leak where the root span is never reported) this will not currently apply.  I'm having difficulty seeing a straightforward solution to that case atm.

johnbley:
> In cases where the `PARTIAL_FLUSH_MIN_SPANS` functionality is triggered (scope leak where the root span is never reported) this will not currently apply. I'm having difficulty seeing a straightforward solution to that case atm.

Good catch.  I'll spend a bit of time digging into that corner case and see what I can come up with.

dd-trace-ot/src/main/java/datadog/opentracing/PendingTrace.java
```diff
@@ -228,6 +229,13 @@ private void expireReference() {
   private synchronized void write() {
     if (isWritten.compareAndSet(false, true)) {
       removePendingTrace();
+      if (size() > Config.get().getMaxSpansPerTrace()) {
+        log.error(
+            "Not writing trace of {} spans, exceeds configured max of {}",
```
tigrannajaryan: Does "not writing" mean dropping? It is not quite clear. Would it make sense to rephrase?

johnbley: Good idea; I will change the message.

johnbley:
Managed a fairly clean way to track and account for partially flushed traces.

dd-trace-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -117,6 +117,9 @@
   public static final String LANGUAGE_TAG_KEY = "language";
   public static final String LANGUAGE_TAG_VALUE = "jvm";
 
+  public static final String MAX_SPANS_PER_TRACE = "maxSpansPerTrace";
```
rmfitzpatrick: The pattern has been dot-separating terms: `max.spans.per.trace`.  This allows configuring w/ `SIGNALFX_MAX_SPAN_PER_TRACE` env var.   Would you mind adding to the configuration table in the main readme as well?

johnbley: Will do both of these once I get tests passing 😄

dd-trace-ot/src/main/java/datadog/opentracing/PendingTrace.java
```diff
@@ -202,9 +205,12 @@ private void expireReference() {
     if (count == 0) {
       write();
     } else {
-      if (tracer.getPartialFlushMinSpans() > 0 && size() > tracer.getPartialFlushMinSpans()) {
+      if (tracer.getPartialFlushMinSpans() > 0
+          && size() > tracer.getPartialFlushMinSpans()
+          && size() + partiallyWrittenSpanCount.get() < Config.get().getMaxSpansPerTrace()) {
```
rmfitzpatrick: Depending on their size and rate of accumulation, this will reveal jumbo traces as leaks as opposed to flushing them as invalid ones.  Is this what we want, or would it be better to move the partially written count informed check below to bypass the `tracer.write()`, still decrementing the completed span count?

johnbley: Excellent point.  I will look at it in more depth, again after I feel like the tests are happier/less flaky.

README.md
```diff
@@ -124,6 +124,7 @@ content](https://docs.signalfx.com/en/latest/apm/apm-overview/apm-metadata.html)
 | `signalfx.db.statement.max.length` | `SIGNALFX_DB_STATEMENT_MAX_LENGTH` | `1024` | The maximum number of characters written for the OpenTracing `db.statement` tag. |
 | `signalfx.recorded.value.max.length` | `SIGNALFX_RECORDED_VALUE_MAX_LENGTH` | `12288` | The maximum number of characters for any Zipkin-encoded tagged or logged value. |
 | `signalfx.trace.annotated.method.blacklist` | `SIGNALFX_TRACE_ANNOTATED_METHOD_BLACKLIST` | `null` | Prevent `@Trace` annotation functionality for the target method string of format `package.OuterClass[methodOne,methodTwo];other.package.OuterClass$InnerClass[*];` (`;` is required and `*` for all methods in class). |
+| `signalfx.max.spans.per.trace` | `SIGNALFX_MAX_SPANS_PER_TRACE` | `2000` | Drop (don't deliver) spans with more traces than this. Intended to prevent runaway traces from flooding upstream systems. |
```
rmfitzpatrick: traces <-> spans

johnbley: Nice catch, fixed.

PR #83, opened 2020-03-22
Make maxSpansPerTrace optional

johnbley:
Set a default value of 0 (meaning "no limit"), add test ensuring that the default is in fact no limit.

### DIFF
--- a/dd-trace-ot/src/main/java/datadog/opentracing/DDTracer.java
+++ b/dd-trace-ot/src/main/java/datadog/opentracing/DDTracer.java
@@ -93,6 +93,8 @@ public class DDTracer implements io.opentracing.Tracer, Closeable, datadog.trace
   private final HttpCodec.Injector injector;
   private final HttpCodec.Extractor extractor;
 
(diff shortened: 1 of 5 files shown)


dd-trace-ot/src/main/java/datadog/opentracing/DDTracer.java
```diff
@@ -217,6 +219,28 @@ public DDTracer(
...
+    this(
+        serviceName,
+        writer,
+        sampler,
+        localRootSpanTags,
+        defaultSpanTags,
+        serviceNameMappings,
+        taggedHeaders,
+        partialFlushMinSpans,
+        Config.DEFAULT_MAX_SPANS_PER_TRACE);
```
rmfitzpatrick: Where is `Config.get().getMaxSpansPerTrace()` being used?

johnbley: Nice catch, updating to use configured value in all cases.

PR #84, opened 2020-03-24
ErrorAnalyzer: change names/values for error tags

johnbley:
- Don't emit error.msg, error.type or error.stack tags
- Do emit sfx.error.{message, object, kind, stack} tags, with values appropriate to the OpenTracing semantic conventions at https://github.com/opentracing/specification/blob/master/semantic_conventions.md#log-fields-table
- Fix up tests accordingly

### DIFF
--- a/dd-java-agent/instrumentation/jsp-2.3/src/test/groovy/JSPInstrumentationBasicTests.groovy
+++ b/dd-java-agent/instrumentation/jsp-2.3/src/test/groovy/JSPInstrumentationBasicTests.groovy
@@ -320,13 +320,14 @@ class JSPInstrumentationBasicTests extends AgentTestRunner {
             "servlet.context" "/$jspWebappContext"
             "http.status_code" 500
             "error" true
-            "error.type" { String tagExceptionType ->
+            "sfx.error.object" { String tagExceptionType ->
               return tagExceptionType == exceptionClass.getName() || tagExceptionType.contains(exceptionClass.getSimpleName())
             }
-            "error.msg" { String tagErrorMsg ->
+            "sfx.error.message" { String tagErrorMsg ->
               return errorMessageOptional || tagErrorMsg instanceof String
             }
-            "error.stack" String
+            "sfx.error.stack" String
+            "sfx.error.kind" String
             defaultTags()
           }
         }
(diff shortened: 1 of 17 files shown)


dd-java-agent/instrumentation/jsp-2.3/src/test/groovy/JSPInstrumentationBasicTests.groovy
```diff
@@ -342,13 +343,14 @@ class JSPInstrumentationBasicTests extends AgentTestRunner {
             "servlet.context" "/$jspWebappContext"
             "jsp.requestURL" reqUrl
             "error" true
-            "error.type" { String tagExceptionType ->
+            "sfx.error.object" { String tagExceptionType ->
```
rmfitzpatrick: Mind adding the modified header to this file?

dd-trace-ot/src/main/java/datadog/opentracing/DDSpan.java
```diff
@@ -152,8 +152,10 @@ public MutableSpan getLocalRootSpan() {
   public void setErrorMeta(final Throwable error) {
     setError(true);
 
+    final String type = error.getClass().getName();
     setTag(DDTags.ERROR_MSG, error.getMessage());
     setTag(DDTags.ERROR_TYPE, error.getClass().getName());
```
rmfitzpatrick: Did you mean to use `type` here?

rmfitzpatrick:
We'll need to clearly detail this in the next release logs.

owais:
Do we continue to use `error: true` boolean tag or migrate it to `sfx.error = true` as well? IIRC, zipkin sets the `error` tag to a human readable error message. As a result it might not always be a boolean across different languages. @rmfitzpatrick @johnbley @flands

owais:
We can still set `error: <value>` in the final formatter/translation layer when converting to zipkin.

owais:
Actually, sticking to `error` is fine. I'll update all cases where we set `error` to strings in the node package.

rmfitzpatrick:
> Do we continue to use `error: true` boolean tag or migrate it to `sfx.error = true` as well? IIRC, zipkin sets the `error` tag to a human readable error message. As a result it might not always be a boolean across different languages. @rmfitzpatrick @johnbley @flands

Our convention thus far has been OpenTracing's where `error` is a bool.

owais:
@rmfitzpatrick Node package sets error to `"true"` (string) instead of bool. Should this be updated/fixed?

PR #85, opened 2020-04-03
Support for Vert.x RxJava

khadalasfx:
Added instrumentation for AsyncResultSingle. This fixes the issue with reactive handlers causality.

### DIFF
--- a/dd-java-agent/instrumentation/vertx/src/main/java/datadog/trace/instrumentation/vertx/VertxRxInstrumentation.java
+++ b/dd-java-agent/instrumentation/vertx/src/main/java/datadog/trace/instrumentation/vertx/VertxRxInstrumentation.java
@@ -1,0 +1,50 @@
+// Modified by SignalFx
+package datadog.trace.instrumentation.vertx;
+
+import static net.bytebuddy.matcher.ElementMatchers.isInterface;
+import static net.bytebuddy.matcher.ElementMatchers.isMethod;
+import static net.bytebuddy.matcher.ElementMatchers.isStatic;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.not;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+import java.util.HashMap;
+import java.util.Map;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+@AutoService(Instrumenter.class)
+public class VertxRxInstrumentation extends Instrumenter.Default {
+
+  public VertxRxInstrumentation() {
+    super("vertx", "vert.x");
+  }
+
+  @Override
+  public ElementMatcher<? super TypeDescription> typeMatcher() {
+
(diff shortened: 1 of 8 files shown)


dd-java-agent/instrumentation/vertx/src/main/java/datadog/trace/instrumentation/vertx/VertxRxInstrumentation.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {
+      packageName + ".AsyncResultSingleAdvice", packageName + ".AsyncResultHandlerConsumerWrapper"
+    };
+  }
+
+  @Override
+  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+    final Map adviceMap = new HashMap();
```
rmfitzpatrick, line 41: Nit: the pattern has been to use `singletonMap`

dd-java-agent/instrumentation/vertx/src/rxHandlerTests/groovy/VertxRxServerTest.groovy
```diff
@@ -0,0 +1,119 @@
...
+import io.vertx.reactivex.core.Vertx
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import spock.lang.Shared
+
+import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.SUCCESS
+
+class VertxRxServerTest extends AgentTestRunner {
+  static {
+    System.setProperty("dd.integration.jdbc.enabled", "true")
```
rmfitzpatrick, line 16: Not jumping out to me why this is necessary if the instrumentation is included with the test task.

dd-java-agent/instrumentation/vertx/vertx.gradle
```diff
@@ -56,11 +60,19 @@ dependencies {
   handlerTestCompile group: 'io.vertx', name: 'vertx-core', version: '3.1.0'
   handlerTestCompile group: 'io.vertx', name: 'vertx-web', version: '3.0.0'
 
+  rxHandlerTestCompile project(':dd-java-agent:instrumentation:netty-4.0')
```
rmfitzpatrick, line 63: doesn't vertx-core 3.6 use netty 4.1?

dd-java-agent/instrumentation/vertx/src/main/java/datadog/trace/instrumentation/vertx/VertxRxInstrumentation.java
khadalasfx: Fixed

dd-java-agent/instrumentation/vertx/src/rxHandlerTests/groovy/VertxRxServerTest.groovy
khadalasfx: I don't know all details how this flag is used, but it is even necessary in https://github.com/signalfx/signalfx-java-tracing/blob/3f8a80189bf942b9ac32c419cd0741cd8ef5e8cf/dd-java-agent/instrumentation/jdbc/src/test/groovy/JDBCInstrumentationTest.groovy#L30

dd-java-agent/instrumentation/vertx/vertx.gradle
khadalasfx: Good catch! Fixed.

PR #87, opened 2020-04-06
Introduce max.continutation.depth with default of 100 to control runaway recursion in async scope propagation.

johnbley:


dd-trace-ot/src/test/groovy/datadog/opentracing/scopemanager/ScopeManagerTest.groovy
```diff
@@ -274,6 +276,33 @@ class ScopeManagerTest extends DDSpecification {
...
+    ArrayList<ContinuableScope.Continuation> conts = new ArrayList<>()
+    for (int i = 0; i < Config.DEFAULT_MAX_CONTINUATION_DEPTH; i++) {
+      conts.add(continuation)
+      def next = (ContinuableScope) continuation.activate()
+      next.setAsyncPropagation(true)
+      continuation = next.capture()
+    }
+
+    then:
+    continuation != null
```
rmfitzpatrick, line 296: Would it be good to add an additional capture and assert it's null?

johnbley: Fixed, meant to go beyond and assert that it was null.

dd-trace-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -120,6 +120,9 @@
   public static final String MAX_SPANS_PER_TRACE = "max.spans.per.trace";
   public static final Integer DEFAULT_MAX_SPANS_PER_TRACE = 0;
 
+  public static final String MAX_CONTINUATION_DEPTH = "max.continutation.depth";
+  public static final Integer DEFAULT_MAX_CONTINUATION_DEPTH = 100;
```
rmfitzpatrick, line 124: What's the confidence level with this default?  Could a truly runaway trace warrant a greater one to distinguish from complex middleware?

PR #90, opened 2020-04-09
Normalize sql

johnbley:
Regex-based approach to normalizing sql string and numeric literals away (replacing them, per industry standard, with "?").

### DIFF
--- a/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCUtils.java
+++ b/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCUtils.java
@@ -5,9 +5,27 @@ import datadog.trace.bootstrap.ExceptionLogger;
 import java.lang.reflect.Field;
 import java.sql.Connection;
 import java.sql.Statement;
+import java.util.regex.Pattern;
 
 public abstract class JDBCUtils {
   private static Field c3poField = null;
+  private static final String[] SqlNormalizeClauses = {
+    // Numerics
+    "0x[0-9a-fA-F]+", // hex
+    "(?<![\\p{LC}_0-9])[+.-]*[0-9]+[0-9xEe.+-]*", // regular numeric literal
+    // Quoted strings
+    "'((?:''|[^'])*)'", // '-quoted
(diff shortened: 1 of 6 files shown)


dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCUtils.java
```diff
@@ -1,12 +1,31 @@
...
 import java.sql.Connection;
 import java.sql.Statement;
+import java.util.regex.Pattern;
 
 public abstract class JDBCUtils {
   private static Field c3poField = null;
+  private static final String[] SqlNormalizeClauses = {
+    // Numerics
+    "0x[0-9a-fA-F]+", // hex
+    "(?<![\\p{LC}_0-9])[+.-]*[0-9]+[0-9xEe.+-]*", // regular numeric literal
```
rmfitzpatrick, line 15: What's the significance of this lookbehind?

johnbley: It's what prevents FIELD2 identifiers from becoming FIELD?.

dd-java-agent/instrumentation/jdbc/src/test/groovy/SqlNormalizerTest.groovy
```diff
@@ -0,0 +1,64 @@
...
+
+    // Not numbers but could be confused as such
+    "SELECT A + B"                                                             | "SELECT A + B"
+    "SELECT -- comment"                                                        | "SELECT -- comment"
+    "SELECT * FROM TABLE123"                                                   | "SELECT * FROM TABLE123"
+    "SELECT FIELD2 FROM TABLE_123 WHERE X<>7"                                  | "SELECT FIELD2 FROM TABLE_123 WHERE X<>?"
+
+    // Semi-nonsensical almost-numbers to elide or not
+    "SELECT --83--...--8e+76e3E-1"                                             | "SELECT ?"
+    "SELECT DEADBEEF"                                                          | "SELECT DEADBEEF"
```
rmfitzpatrick, line 42: Mind adding checks for `123-45-6789` and `1/2/34` etc?  I think tEsTing CAse insensitivity would also be a good idea.

PR #92, opened 2020-04-10
Significantly reduce thrash allocations in zipkin encoding

johnbley:
* Use a short-circuit for turning 64-bit IDs into hex
* Remove some spurious extra allocations/doubled work
* Rework buffer strategy between serialization and writing to the wire to not double-copy traces.

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/Ids.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/Ids.java
@@ -5,9 +5,31 @@ import java.math.BigInteger;
 
 /** Conversions between DataDog decimal ids and Zipkin hex ids. Supports both 64 and 128 bit ids. */
 public class Ids {
-  private static char[] pad = "00000000000000000000000000000000".toCharArray();
+  private static final char[] pad = "00000000000000000000000000000000".toCharArray();
+  private static final char[] HexChars = {
+    '0', '1', '2', '3',
+    '4', '5', '6', '7',
+    '8', '9', 'a', 'b',
+    'c', 'd', 'e', 'f',
+  };
 
   public static char[] idToHexChars(final String id) {
+    if (id.length() <= 20) {
+      // *likely* (but not definitely) a 64-bit id: try to parse as such
+      try {
+        long val = Long.parseLong(id);
+        char[] answer = new char[16];
+        for (int i = 0; i < 16; i++) {
+          long mask = (0x0FL << (i * 4));
+          long nybble = (val & mask) >> (i * 4);
+
+          answer[15 - i] = HexChars[(int) (nybble & 0xFF)];
+        }
+        return answer;
+      } catch (NumberFormatException thatsFine) {
+        // continue on to using BigInteger below
+      }
+    }
     final String asHex = new BigInteger(id, 10).toString(16);
     final int desiredLength = asHex.length() > 16 ? 32 : 16;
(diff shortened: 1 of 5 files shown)


dd-trace-api/src/main/java/datadog/trace/api/Ids.java
```diff
@@ -5,9 +5,31 @@
...
+  };
 
   public static char[] idToHexChars(final String id) {
+    if (id.length() <= 20) {
+      // *likely* (but not definitely) a 64-bit id: try to parse as such
+      try {
+        long val = Long.parseLong(id);
+        char[] answer = new char[16];
+        for (int i = 0; i < 16; i++) {
+          long mask = (0x0FL << (i * 4));
```
rmfitzpatrick, line 23: these can be precomputed or will the jvm likely do that?

dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
```diff
@@ -67,22 +66,24 @@ public Response sendTraces(final List<List<DDSpan>> traces) {
...
-  public byte[] serializeTrace(final List<DDSpan> trace) throws IOException {
-    final ByteArrayOutputStream stream = new ByteArrayOutputStream(trace.size() * 128);
+  public SerializedBuffer serializeTrace(final List<DDSpan> trace) throws IOException {
+    final StreamingSerializedBuffer stream = new StreamingSerializedBuffer(trace.size() * 128);
     final JsonGenerator jsonGenerator = JSON_FACTORY.createGenerator(stream, JsonEncoding.UTF8);
 
     jsonGenerator.writeStartArray();
-    for (DDSpan span : trace) {
-      writeSpan(span, jsonGenerator);
+    for (int i = 0; i < trace.size(); i++) {
```
rmfitzpatrick, line 74: Is this iterator adding overhead?

dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
```diff
@@ -91,14 +92,14 @@ public Response sendSerializedTraces(
         int traceCount = 0;
 
         out.write('[');
```
rmfitzpatrick, line 94: This has always made me feel a little dirty.  Would using the streaming api be better in your opinion?

dd-trace-api/src/main/java/datadog/trace/api/Ids.java
johnbley: Not worth it; a move is so much more expensive than a one-op shift.  Also, the memory allocation saved here outweighs either by several orders of magnitude.

dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
johnbley: It's a tiny piece of the overall puzzle and almost certainly elided by escape analysis, but every bit matters for this particular allocation hot path.

dd-trace-ot/src/main/java/datadog/trace/common/writer/ZipkinV2Api.java
johnbley: I wouldn't lose any sleep over it; this code is EOL eventually anyway.

PR #96, opened 2020-04-20
Set span.kind=server only for receiving end of a distributed trace

khadalasfx:
Only the local root span can have span.kind=server, because otherwise
it leads to undesired endpoint metrics in PG.

### DIFF
--- a/dd-java-agent/agent-tooling/src/test/groovy/datadog/trace/agent/decorator/BaseDecoratorTest.groovy
+++ b/dd-java-agent/agent-tooling/src/test/groovy/datadog/trace/agent/decorator/BaseDecoratorTest.groovy
@@ -1,3 +1,4 @@
+// Modified by SignalFx
 package datadog.trace.agent.decorator
 
 import datadog.trace.agent.test.utils.ConfigUtils
--- a/dd-java-agent/agent-tooling/src/test/groovy/datadog/trace/agent/decorator/BaseDecoratorTest.groovy
+++ b/dd-java-agent/agent-tooling/src/test/groovy/datadog/trace/agent/decorator/BaseDecoratorTest.groovy
@@ -23,6 +23,7 @@ class BaseDecoratorTest extends DDSpecification {
     1 * span.setTag(DDTags.SPAN_TYPE, decorator.spanType())
     1 * span.setTag(Tags.COMPONENT.key, "test-component")
     _ * span.setTag(_, _) // Want to allow other calls from child implementations.
+    _ * span.getLocalRootSpan() // Want to allow other calls from child implementations.
     0 * _
   }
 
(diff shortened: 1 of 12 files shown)


dd-java-agent/instrumentation/ratpack-1.4/src/test/groovy/server/RatpackHttpServerTest.groovy
```diff
@@ -113,7 +113,6 @@ class RatpackHttpServerTest extends HttpServerTest<EmbeddedApp, NettyHttpServerD
         "$Tags.PEER_PORT" Integer
         "$Tags.PEER_HOST_IPV4" { it == null || it == "127.0.0.1" } // Optional
         "$Tags.HTTP_METHOD" String
-        "$Tags.SPAN_KIND" Tags.SPAN_KIND_SERVER
```
rmfitzpatrick, line 116: Mind adding a modified header?

dd-java-agent/instrumentation/play-2.4/src/test/groovy/server/PlayServerTest.groovy
```diff
@@ -92,7 +92,6 @@ class PlayServerTest extends HttpServerTest<Server, NettyHttpServerDecorator> {
         "$Tags.HTTP_URL" String
         "$Tags.PEER_HOST_IPV4" { it == null || it == "127.0.0.1" } // Optional
         "$Tags.HTTP_METHOD" String
-        "$Tags.SPAN_KIND" Tags.SPAN_KIND_SERVER
```
rmfitzpatrick, line 95: same request

dd-java-agent/instrumentation/ratpack-1.4/src/test/groovy/server/RatpackHttpServerTest.groovy
khadalasfx: Sorry, I forgot about this - fixed!

dd-java-agent/instrumentation/play-2.4/src/test/groovy/server/PlayServerTest.groovy
khadalasfx: done

dd-java-agent/agent-tooling/src/test/groovy/datadog/trace/agent/decorator/BaseDecoratorTest.groovy
```diff
@@ -23,6 +24,7 @@ class BaseDecoratorTest extends DDSpecification {
     1 * span.setTag(DDTags.SPAN_TYPE, decorator.spanType())
     1 * span.setTag(Tags.COMPONENT.key, "test-component")
     _ * span.setTag(_, _) // Want to allow other calls from child implementations.
+    _ * span.getLocalRootSpan() // Want to allow other calls from child implementations.
     0 * _
   }
```
johnbley, line 30: One idea to better ensure all the paths that produce kind=server spans have been fixed is to go into TraceAssert.assertTrace and walk the spans in the trace, asserting that at most one has kind=server.  This would get applied to all the traces in the test suite.

khadalasfx: I had to add this "_ * span.getLocalRootSpan()" to allow ServerDecoratorTest to work, it actually not testing a new behavior. 

Most of the tests validate "span.kind" (that's why so many test files I have to update) but I really like your idea - if some instrumentation doesn't validate that, it is good place to catch it.

I've added an extra check, in TraceAssert.assertTrace as you suggested. Thanks.

PR #101, opened 2020-04-25
Introduce new lexer-based sql normalization.

johnbley:
This improves performance quite a bit in the general case and, more importantly, fixes several edge cases in regex handling like stack overflows and very slow loopbacks.
Adds tests showing that these edge cases don't affect the lexer implementation.

Fixes a subtle muzzle bug involving primitive types and helper classes that was in the way of this work.

Adjusts decorators to not emit the db.statement tag if normalization fails (which should be quite rare).


dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/muzzle/ReferenceMatcher.java
```diff
@@ -183,15 +184,31 @@ public boolean matches(final ClassLoader loader) {
     return mismatches;
   }
 
+  private static boolean matchesPrimitive(String longName, String shortName) {
+    // The two meta type systems in use here differ in their treatment of primitive type names....
```
iNikem, line 188: Which two type systems?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCUtils.java
```diff
@@ -71,8 +53,12 @@ public static Connection connectionFromStatement(final Statement statement) {
...
   }
 
+  /** @return null if the sql can't be normalized properly (should be quite rare) */
   public static String normalizeSql(String sql) {
-    // FIXME performance could be replaced by a much more efficient lexer.
-    return SqlNormalizePattern.matcher(sql).replaceAll("?");
+    try {
+      return SqlNormalizer.normalize(sql);
+    } catch (Exception e) {
+      return null;
```
iNikem, line 61: Log exception?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -44,6 +44,14 @@ public StatementInstrumentation() {
   public String[] helperClassNames() {
     final List<String> helpers = new ArrayList<>(JDBCConnectionUrlParser.values().length + 9);
 
+    helpers.add(packageName + ".normalizer.Token");
```
iNikem, line 47: This code is duplicated 3 times. Can it be extracted and reused?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/normalizer/SqlNormalizer.jj
```diff
@@ -0,0 +1,184 @@
...
+	{
+		return sb;
+	}
+}
+
+TOKEN :
+{
+// Javacc doesn't support unicode character categories like {Letter}, so here's my approximation of that.  No, I didn't type this by hand.
+
+	< Id: [
```
iNikem, line 96: I am not sure I understand what does this "Id" represent :) Can you add a comment, please?

iNikem:
Do you have any performance tests to add to demonstrate the improvement?

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/muzzle/ReferenceMatcher.java
johnbley: One is from ByteBuddy and the other is from the otel auto agent.  Their APIs are so similar and this code is so annoying because of it.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCUtils.java
johnbley: Good catch, will do.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
johnbley: None of the other decorators do this, so out of muzzle superstition I follow their pattern.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/normalizer/SqlNormalizer.jj
johnbley: I could come up with a better name certainly - it's meant to represent a keyword (like SELECT) or identifier (like MY_SPECIAL_TABLE2).  Will change to KeywordOrIdentifier and a comment based on this explanation :-)

johnbley:
> I did not review generated files. Do we need some instructions for future developers about how to make changes in the lexer?

I put some instructions for that in the top of the .jj file.

johnbley:
> Do you have any performance tests to add to demonstrate the improvement?

I did - tests like the very long numbers and the lots of ticks tests were failing before with the regex -based implementation.  In time I'll add more.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/normalizer/SqlNormalizer.jj
```diff
@@ -0,0 +1,191 @@
...
+	)* <EOF>
+
+	{
+		return sb;
+	}
+}
+
+TOKEN :
+{
+// Javacc doesn't support unicode character categories like {Letter}, so here's my approximation of that.  No, I didn't type this by hand.
```
rmfitzpatrick, line 94: Can you provide a snippet or description of how this was generated and why so many disparate ranges are needed?  Why not fall under Other?

johnbley: Will do, give me a sec.

PR #107, opened 2020-05-12
Only set tracing library tags on local root

rmfitzpatrick:
These changes move the version information from default global tags to those of the local root.


dd-trace-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -579,6 +575,8 @@ private Config(final Properties properties, final Config parent) {
   public Map<String, String> getLocalRootSpanTags() {
     final Map<String, String> runtimeTags = getRuntimeTags();
     final Map<String, String> result = new HashMap<>(runtimeTags);
+    result.put(TRACING_LIBRARY_KEY, TRACING_LIBRARY_VALUE);
+    result.put(TRACING_VERSION_KEY, TRACING_VERSION_VALUE);
```
johnbley, line 580: This was a lot cleaner than I expected; nice!

PR #118, opened 2020-07-20
[APMI-94] set JDK tests for all (7-14) versions to ZULU, cleaned-up naming a bit

ghost:
Task: https://signalfuse.atlassian.net/browse/APMI-94
Discussion: https://splunk.slack.com/archives/C011XR5KLKX/p1594904868057700

Changes:
- cleaned-up duplicated tests for same version
- unified to ZULU (apart from latest dep test)
- changed naming to that it's always clear which JDK is being used

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,54 +86,49 @@ jobs:
       - store_test_results:
           path: ./results
 
-  test_7:
+  test_zulu_7:
     <<: *default_test_job
     environment:
       - TEST_TASK: testJava7
 
-  test_8:
+  test_zulu_8:
     <<: *default_test_job
     environment:
-      # We are building on Java8, this is our default JVM so no need to set more homes
-      - TEST_TASK: test jacocoTestReport jacocoTestCoverageVerification
+      # default JDK provider is ZULU, hence home set
+      - TEST_TASK: testJavaZULU8 jacocoTestReport jacocoTestCoverageVerification
 
-  test_latest8:
+  test_latest_8:
     <<: *default_test_job
     environment:
       # We are building on Java8, this is our default JVM so no need to set more homes
       - TEST_TASK: latestDepTest
 
-  test_ibm8:
+  test_zulu_9:
     <<: *default_test_job
     environment:
-      - TEST_TASK: testJavaIBM8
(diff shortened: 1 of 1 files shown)


johnbley (approved):

.circleci/config.yml
```diff
@@ -86,54 +86,49 @@ jobs:
...
     <<: *default_test_job
     environment:
-      - TEST_TASK: testJava11
+      - TEST_TASK: testJava10
 
-  test_zulu11:
+  test_zulu_11:
     <<: *default_test_job
     environment:
       - TEST_TASK: testJavaZULU11
```
iNikem, line 119: Why is this different from others? `testJava10` but `testJavaZULU11`?

.circleci/config.yml
```diff
@@ -86,54 +86,49 @@ jobs:
...
     environment:
       - TEST_TASK: testJava7
 
-  test_8:
+  test_zulu_8:
     <<: *default_test_job
     environment:
-      # We are building on Java8, this is our default JVM so no need to set more homes
-      - TEST_TASK: test jacocoTestReport jacocoTestCoverageVerification
+      # default JDK provider is ZULU, hence home set
```
iNikem, line 97: I think this comment is very confusing here. What home? If Zulu is default provider, why `testJavaZULU8` and not just `testJava8` or `test`?

iNikem (approved):

.circleci/config.yml
kubawach: It's confusing I admit but stems directly from how our gradle build creates test tasks. Names of the tasks are created out of Docker env variables and creator of the image (data dog employee) decided to name some ZULU JDKs with "zulu" in the name and some without (8 and 11 have zulu in name). To help avoid the confusion I named jobs accordingly (with zulu in the name added everywhere). To make it better we'd need to change the gradle build as well but I didn't want to modify too much. Hope that makes more sense now :)

.circleci/config.yml
kubawach: Same as the other question - due to env vars naming confusion (zulu, openJdk, ...) I decided to put name of the provider with each job to make it clear which JDK is being used (while task names are autogenerated from env vars).

PR #119, opened 2020-07-20
Add a separate HandlerMappingInstrumentation to update server span name when the corresponding handler got selected

iNikem:
This works around the issue in SWAT-1755 ticket, where span name was not updated if some `HandlerInterceptor` throws an exception from `preHandle`


iNikem:
Test currently fails, I am still investigating

PR #123, opened 2020-08-27
Avoid doubly tracing MongoClientSettingsBuilder

### LINKED ISSUE #122: Mongo instrumentation may duplicate traced command spans
We have observed Mongo queries being traced twice: Each duplicate span has a different span ID but the same parent span, and all span tags are the same. Investigating the issue, we found this issue filed against a similar project:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/457

This issue describes several scenarios where duplicate mongo `CommandListener`s could be injected into a `MongoClientSettings$Builder` instance. This problem appears to affect this project as well, as the mechanism for injecting a `CommandListener` is very similar to the approach used by open-telemetry/opentelemetry-java-instrumentation.

teleological:
Fixes #122: Mongo instrumentation may duplicate traced command spans

Check for already injected TracingCommandListener before injecting.
Patch derived from the fix for a similar issue in open-telemetry/opentelemetry-java-instrumentation.

### DIFF
--- a/dd-java-agent/instrumentation/mongo/driver-3.1/src/main/java/datadog/trace/instrumentation/mongo/MongoClientInstrumentation.java
+++ b/dd-java-agent/instrumentation/mongo/driver-3.1/src/main/java/datadog/trace/instrumentation/mongo/MongoClientInstrumentation.java
@@ -1,6 +1,7 @@
 package datadog.trace.instrumentation.mongo;
 
 import static java.util.Collections.singletonMap;
+import static net.bytebuddy.matcher.ElementMatchers.declaresField;
 import static net.bytebuddy.matcher.ElementMatchers.declaresMethod;
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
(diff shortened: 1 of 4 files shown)


rmfitzpatrick:
@teleological, thank you for your fix.  Would you mind pushing an amended commit to try to engage the Circle job?  There was a setting issue I've corrected--though we may need to close and reopen :(

teleological:
Thanks for looking at this, @rmfitzpatrick. I amended and re-pushed and it looks like Circle is building.

johnbley:
Hi @teleological - it looks like you'll need to run `./gradlew googleJavaFormat` to get the source into the proper format for our CI.  Some of the tests have sporadic failures - I'll worry about restarting those as necessary.

teleological:
Thank you, @johnbley: On it!

teleological:
Formatted. Glad to squash. LMK

johnbley:
> Formatted. Glad to squash. LMK

No need; will squash on merge.  Thanks again!

PR #125, opened 2020-08-31
Fix #124: Enable priority samplers to drop traces

### LINKED ISSUE #124: No way to drop traces in Java
SignalFX's tracing library for Java, unlike those for Ruby and Python, doesn't appear to provide a way to exclude traces from being forwarded. The Ruby and Python libraries, by comparison, make use of Jaeger clients which use the OpenTracing `sampling.priority` tag to determine whether to forward a trace.

The forked client in the Java library doesn't offer explicit support for `sampling.priority`, although it appears there may be server-side support for using it to force trace retention: https://docs.signalfx.com/en/latest/apm-pg/apm-pg-deployment/advanced-gateway-features.html#force-trace-retention (This feature may not be available to Ruby/Python/etc since the tag is intercepted by the client.)

On the other hand, the Java library offers some support for `manual.drop` and `manual.keep` tags, which appear as if they should enable this functionality. However, it seems that all of the configurable Samplers ignore span priority,  whether configured by their rules or by these tags, and always tell the Tracer to forward spans.

It's unclear whether the metrics property which is set according to these tags when a priority sampler is enabled (`_sampling_priority_v1`) causes traces to receive different processing once they have been received by SignalFX. I was able to confirm that `manual.drop` which sets the value for this metric to -1, does not have the effect of excluding traces from being recorded.

teleological:
This patch ensures that Rate-By-Service and Rule-Based samplers drop traces as configured, or when spans set `manual.drop` or `manual.keep`. The behavior of the default "All" Sampler is not changed.

### DIFF
--- a/dd-trace-ot/src/main/java/datadog/trace/common/sampling/RateByServiceSampler.java
+++ b/dd-trace-ot/src/main/java/datadog/trace/common/sampling/RateByServiceSampler.java
@@ -30,9 +30,8 @@ public class RateByServiceSampler implements Sampler, PrioritySampler, ResponseL
 
   @Override
   public boolean sample(final DDSpan span) {
-    // Priority sampling sends all traces to the core agent, including traces marked dropped.
-    // This allows the core agent to collect stats on all traces.
-    return true;
+    final Integer samplingPriority = span.getSamplingPriority();
+    return (samplingPriority != null && samplingPriority > 0);
   }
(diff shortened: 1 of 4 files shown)


johnbley:
See comment in #124  - please evaluate the OpenTelemetry agent (and possibly make a similar contribution there).  Thanks!

PR #126, opened 2020-09-24
camel instrumentation

ghost:
Resolves APMI-1057

In order to meet Nike's requirements for improved Camel tracing in SFx agent (Camel provides separate OTEL module):
- integrated https://github.com/apache/camel/tree/camel-2.20.x/components/camel-opentracing
- changed the code to make it work with SFx autoinstrumentation
- added guards (multiple instrumentation etc.)
- fixed OpenTracing / agent integration (keeping spans from different instrumentations in sync)

"Business" side of the instrumentation remained untouched!

Tested with sample Camel 2.20.x applications (client, two services). Trace ID propagation confirmed 👍 

Current state:

- integration tests added
- traces generated properly (child/parent, trace id propagation)
- issue with Servlet3 instrumentation when "jetty" Camel route is used - Servlet3 instrumentation adds extra span ("org.eclipse.jetty.servlet.ServletHandler$CachedChain") as Camel calls distpach() in the middle of the filer chain, which causes Serlvet instrumentation to close existing server span  

NOTE:
- JDK 7 test failures (not relevant - Camel is 8 up)
- JDK 8 latest deps failures - not relevant as caused by Mongo tests that fail master as well


iNikem:
@kubawach can you point out where do you expect for those separate spans to be in the bigger trace?

kubawach:
> @kubawach can you point out where do you expect for those separate spans to be in the bigger trace?

Happy to do so! In the provided example, there are following issues:
- two traces - 04f0ca77faa19115 (started by commons-http-client) and 46a39b709d6e0ea1 (started by camel-timer - the client call)
- the second trace contains only two spans - client (main timer) and client (http component executing the call)
- the first one contains spans for all three tiers (client, service 1, service 2) but spans fork each time camel is involved, for example spans 3d68d0ae8a7ab1a9 and 6d467e5de5adc6ce have own branches, for camel and other instrumentations respectively, even though there is a single execution
- single span from the third layer (SERVICE 2) - 628a3334cd8da8f7 - has "http client" span from SERVICE 1 (711c68ab882e494e) as the parent

dd-java-agent/instrumentation/apache-camel-2.20/src/main/java/org/apache/camel/opentracing/OpenTracingTracer.java
```diff
@@ -0,0 +1,265 @@
...
+
+    @Override
+    public void onExchangeBegin(Route route, Exchange exchange) {
+      try {
+
+        SpanDecorator sd = getSpanDecorator(route.getEndpoint());
+        Span span =
+            tracer
+                .buildSpan(sd.getOperationName(exchange, route.getEndpoint()))
+                .asChildOf(
```
iNikem, line 226: I think here we first have to check if there is active "our" span already present. If yes, then use it as parent, no remote span extracted from the wire

dd-java-agent/instrumentation/apache-camel-2.20/src/main/java/org/apache/camel/opentracing/ActiveSpanManager.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /**
+   * This method activates the supplied span for the supplied exchange. If an existing span is found
+   * for the exchange, this will be pushed onto a stack.
+   *
+   * @param exchange The exchange
+   * @param span The span
+   */
+  public static void activate(Exchange exchange, Span span) {
```
iNikem, line 48: I would try to call `datadog.trace.bootstrap.instrumentation.api.AgentTracer#activateSpan` here to active our span at the same time as Camel actives its span. I don't quite understand when and how we should close the resulting scope.

iNikem:
@kubawach First, I think the second trace is not important here at all. It is just a second trace irrelevant to the first one. But that is not the most important part :)

It seems that indeed we have problems because span manipulations done by camel and by our agent are not in sync. I have left a couple of comments in places where, I think, we should try to synchronise two systems. Give it a try.

dd-java-agent/instrumentation/apache-camel-2.20/src/main/java/org/apache/camel/opentracing/OpenTracingTracer.java
kubawach: Done.

dd-java-agent/instrumentation/apache-camel-2.20/src/main/java/org/apache/camel/opentracing/ActiveSpanManager.java
kubawach: Done.

iNikem (approved):

PR #135, opened 2020-12-22
🌶️  Update to latest jmxfetch upstream 🌶️

breedx-splk:
...and mitigate transitive jackson vulns.

Noting this as spicy, since [quite a few changes were made in jmxfetch](https://github.com/DataDog/jmxfetch/blob/master/CHANGELOG.md) since `0.35.0`.

Upstream lost a number of dependencies, including `jackson-databind` and `commons-io` (and several others).  As a result, we have to now depend on `commons-io` directly (for IOUtils).

### DIFF
--- a/dd-java-agent/agent-jmxfetch/src/main/java/datadog/trace/agent/jmxfetch/JMXFetch.java
+++ b/dd-java-agent/agent-jmxfetch/src/main/java/datadog/trace/agent/jmxfetch/JMXFetch.java
@@ -2,18 +2,13 @@ package datadog.trace.agent.jmxfetch;
 
 import static org.datadog.jmxfetch.AppConfig.ACTION_COLLECT;
 
-import com.google.common.collect.ImmutableList;
 import datadog.trace.api.Config;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.SortedSet;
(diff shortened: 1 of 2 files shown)


dd-java-agent/agent-jmxfetch/src/main/java/datadog/trace/agent/jmxfetch/JMXFetch.java
```diff
@@ -69,7 +64,7 @@ private static void run(final Config config) {
 
     final AppConfig.AppConfigBuilder configBuilder =
         AppConfig.builder()
-            .action(ImmutableList.of(ACTION_COLLECT))
+            .action(Collections.unmodifiableList(Arrays.asList(ACTION_COLLECT)))
```
jkwatson, line 67: why not just `Collections.singletonList()` ?

dd-java-agent/agent-jmxfetch/src/main/java/datadog/trace/agent/jmxfetch/JMXFetch.java
```diff
@@ -23,8 +18,8 @@
 @Slf4j
 public class JMXFetch {
 
-  public static final ImmutableList<String> DEFAULT_CONFIGS =
-      ImmutableList.of("jmxfetch-config.yaml");
+  public static final List<String> DEFAULT_CONFIGS =
+      Collections.unmodifiableList(Arrays.asList("jmxfetch-config.yaml"));
```
jkwatson, line 22: Why not just `Collections.singletonList()` here?

jkwatson (approved):

PR #137, opened 2021-02-26
Expose service name and environment via MDC

seemk:
* New system property / env var: `SIGNALFX_ENVIRONMENT_NAME`. `default: ""`.
* If environment name is set, it is added to span tags as `environment`
* Expose both service name and environment to loggers similar to how trace and span IDs are exposed

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ over corresponding environment variables.
 | System property | Environment variable | Default value | Notes |
 | ---             | ---                  | ---           | ---   |
 | `signalfx.service.name` | `SIGNALFX_SERVICE_NAME` | `"unnamed-java-service"` | The name of the service. |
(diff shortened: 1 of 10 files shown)


iNikem (approved):

seemk:
Note: changing the environment variable name to `SIGNALFX_ENV`.

johnbley (approved):

johnbley:
LGTM, let me know when you want to turn a release and I can handle that in my Copious Spare Time.

seemk:
Don't / won't merge yet - changing the configuration options a bit.

seemk:
Replaced `SIGNALFX_ENVIRONMENT_NAME` with `SIGNALFX_ENV` to be more consistent with other agents. `ENV` already existed before, but only as an env var, so fixed this.

PR #141, opened 2021-12-10
Upgrade log4j2 to latest 2.15.0

breedx-splk:
Even though it's just the `testCompile`, I assume this would get flagged at some point so let's upgrade it.


breedx-splk:
Superseded by #143

